PR #2, opened 2022-02-01
Feature: Simplify base props

oleg-semyonov:
- Removed properties unused by `NewsroomContextProvider`
- Simplified types by re-using `NewsroomContextProps` type
- Renamed `BasePageProps` type to `NewsroomContextProps` to better represent its function
- Renamed `getBasePageProps` to `getNewsroomServerSideProps` to better represent its function
- Condensed props for `NewsroomContextProvider` into a single property to reduce boilerplate
- Added `npm run release:preview` script to publish preview versions of the package


src/adapter-nextjs/processRequest.ts
```diff
@@ -1,31 +1,52 @@
...
     context: GetServerSidePropsContext,
-    basePageProps: BasePageProps,
+    props: Props & PageProps & ServerSidePageProps,
     canonicalUrl?: string,
-    customProps?: Omit<Props, keyof BasePageProps>,
 ): GetServerSidePropsResult<Props> {
-    const { locale, query } = context;
+    const { locale: nextLocale, query } = context;
+
+    const { pageProps, serverSideProps } = extractServerSideProps(props);
```
kudlajz, line 27: What's the point of this function? Can't you just do
```typescript
const { localeResolved, ...pageProps } = props;

if (!localeResolved) {
    return { notFound: true };
}

if (canonicalUrl) {
    const { languages, localeCode } = pageProps.newsroomContextProps;
    const currentLocale = LocaleObject.fromAnyCode(localeCode);
    const shortestLocaleCode = getShortestLocaleCode(languages, currentLocale);

    const redirect = getRedirectToCanonicalLocale(
        shortestLocaleCode,
        nextLocale,
        canonicalUrl,
        query,
    );
    if (redirect) {
        return { redirect };
    }
}

return {
    props: pageProps as Props,
};
```

This way `localeResolved` will be dropped from the `pageProps` and then you just return them. Typescript doesn't complain so it should work?

kudlajz (commented):
Looks good 👍  I have one thing which would simplify it a bit more.

src/adapter-nextjs/processRequest.ts
oleg-semyonov: This is a bit of premature refactor, initially it was the same way as in your suggestion, but the point is, if we change those types in the future, there will be no warnings or error from TS, which is not optimal. The function helps to check at least the `ServerSidePageProps` type.
But maybe I'm overthinking this :D

kudlajz: Well the worst thing that could happen would be that some of the server-side props would be passed through to the component, right?

So I'd say it sounds like overthinking.. 😄

oleg-semyonov: You're right, although this would be hard to spot and debug if you don't know where to look at. But let's leave it for future us :D

oleg-semyonov: Oh, the unknown part actually fixed itself somehow. I guess it was due to switch from `extends` interface to an union type. So it's all good now!

kudlajz (approved):

PR #13, opened 2022-03-14
[TITS-5216] Added IS_PREVIEW env variable

kudlajz:
Added `IS_PREVIEW` environment variable to provide themes with information on whether tracking should be enabled/disabled and or any other behaviour changes based on this.

### DIFF
--- a/src/data-fetching/types.ts
+++ b/src/data-fetching/types.ts
@@ -5,6 +5,7 @@ export interface AlgoliaSettings {
 }
 
 export interface PrezlyNewsroomEnv {
+    IS_PREVIEW: string;
     PREZLY_ACCESS_TOKEN: string;
     PREZLY_NEWSROOM_UUID: string;
     PREZLY_THEME_UUID?: string;


src/data-fetching/types.ts
```diff
@@ -5,6 +5,7 @@ export interface AlgoliaSettings {
 }
 
 export interface PrezlyNewsroomEnv {
+    IS_PREVIEW: string;
```
kudlajz, line 8: Is it enough to just add it here? It should be as long as it's provided via deployment right?

oleg-semyonov: If you want to use this variable on the client side, you'll need to prefix it with `NEXT_PUBLIC_`.
Otherwise, should be good. But better to make it optional.

oleg-semyonov (changes requested):
I think the variable should be marked as optional, there is no need to force it on developers using the library, when it's really mostly for our internal use.

src/data-fetching/types.ts
kudlajz: I see. Should we then maybe add it next to https://github.com/prezly/theme-nextjs-bea/blob/main/@types/global.d.ts#L8 instead of here? 🤔

PR #16, opened 2022-03-22
[TITS-5301] Fix next adapter not working for single-language apps

oleg-semyonov:
These changes allow for themes that don't support multi-language to still work with theme-kit.
Previously, if `i18n` section of `next.config.js` was not setup for multi-language support, theme would endlessly redirect to the same locale.

Tested on the Heartbeat theme, where this issue was first discovered.
Also tested on Bea to confirm that regular multi-language behavior is still working as intended

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prezly/theme-kit-nextjs",
-  "version": "1.6.2",
+  "version": "1.6.3-0",
   "description": "Data layer and utility library for developing Prezly themes with NextJS",
   "main": "build/index.js",
   "types": "build/index.d.ts",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@prezly/theme-kit-nextjs",
-  "version": "1.6.2",
+  "version": "1.6.3-0",
```
kudlajz, line 3: Should this be committed? 🤔

kudlajz (approved):

package.json
oleg-semyonov: This was auto-pushed when publishing a release preview for testing. I think it's fine to commit it, since there is a package with this version on NPM

PR #17, opened 2022-03-22
Add return type override to `getEnvVariables`

oleg-semyonov:
This feature was done mainly because of [this code in the Heartbeat Theme](https://github.com/prezly/theme-nextjs-heartbeat/blob/7a47166c4fa880849e45d080f41042d6f4869e95/pages/%5Bslug%5D.tsx#L41-L46).

The thing is that it'd be nice to use the same `getEnvVariables` method provided by theme-kit to also get any custom env variables specific to the theme (which it does either way, but TS is complaining about type mismatch).

Now I'm not sure whether the type cast added here is better than just doing this in the Heartbeat code:
```ts
const env = getEnvVariables(context.req) as ProcessEnv;
```

### DIFF
--- a/src/data-fetching/lib/getEnvVariables.ts
+++ b/src/data-fetching/lib/getEnvVariables.ts
@@ -29,19 +29,18 @@ function decodeHttpEnv(header: string): Record<string, any> {
 
 export function getEnvVariables<EnvType extends PrezlyEnv = PrezlyEnv>(
     req?: IncomingMessage,
-): EnvType {
+): NodeJS.ProcessEnv & EnvType {
     if (typeof window !== 'undefined') {
         throw new Error('"getEnvVariables" should only be used on back-end side.');
     }
 
     const headerName = (process.env.HTTP_ENV_HEADER || '').toLowerCase();
-    const processEnv = process.env as unknown as EnvType;
 
     if (headerName && req) {
         const header = req.headers[headerName] as string | undefined;
         const httpEnv = decodeHttpEnv(header || '');
-        return { ...processEnv, ...httpEnv };
+        return { ...process.env, ...httpEnv } as NodeJS.ProcessEnv & EnvType;
     }
 
-    return { ...processEnv };
+    return { ...process.env } as NodeJS.ProcessEnv & EnvType;
 }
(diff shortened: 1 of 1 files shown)


kudlajz (approved):
I gave it a long thought but couldn't come up with a better solution.

The main problem is that sometimes `PrezlyEnv` properties are part of `process.env` (when running locally) and other times they are taken from the HTTP header (when running on Prezly infrastructure).

So `process.env` has actually two possible shapes based on where it's running.

I don't see a way to type that at the moment.

src/data-fetching/lib/getEnvVariables.ts
```diff
@@ -27,18 +27,21 @@ function decodeHttpEnv(header: string): Record<string, any> {
     return decodeJson(header);
 }
 
-export function getEnvVariables(req?: IncomingMessage): PrezlyEnv {
+export function getEnvVariables<EnvType extends PrezlyEnv = PrezlyEnv>(
```
kudlajz, line 30: Maybe it could look like this? 🤔  So we at least show that we're returning both the `EnvType` and `ProcessEnv`.
With your implementation it seems like we're only returning `EnvType` which is not entirely correct.

```typescript
export function getEnvVariables<EnvType extends PrezlyEnv = PrezlyEnv>(
    req?: IncomingMessage,
): NodeJS.ProcessEnv & EnvType {
    if (typeof window !== 'undefined') {
        throw new Error('"getEnvVariables" should only be used on back-end side.');
    }

    const headerName = (process.env.HTTP_ENV_HEADER || '').toLowerCase();

    if (headerName && req) {
        const header = req.headers[headerName] as string | undefined;
        const httpEnv = decodeHttpEnv(header || '');
        return { ...process.env, ...httpEnv } as NodeJS.ProcessEnv & EnvType;
    }

    return { ...process.env } as NodeJS.ProcessEnv & EnvType;
}

oleg-semyonov: Yes, that would make more sense! Thanks for looking into this :)

kudlajz: I was thinking about adding a check whether the HTTP env header exists (and throwing error if it doesn't) but that doesn't work locally.

PR #18, opened 2022-03-22
[TITS-5308] Extract useInfiniteLoading hook

kudlajz:


package.json
```diff
@@ -45,7 +45,8 @@
   "peerDependencies": {
     "next": "^12.x",
     "react": "^17.x",
-    "react-dom": "^17.x"
+    "react-dom": "^17.x",
+    "react-use": "^17.3.x"
```
oleg-semyonov, line 49: I'm not a big fan of bringing a new peer dependency for a single hook 😕 
Can we import just the `useAsyncFn` hook from the library? Then we can make it a direct dependency.
`react-use` is really nice, but I don't feel that everybody would be happy to see it being a requirement for this library.

oleg-semyonov (commented):
Looks good, although I'm not sure about the new peer dependency 🤔

package.json
kudlajz: How would that work? It's not available as a standalone package 🤔

oleg-semyonov: Oh, I was thinking it works similarly to `lodash` sub-packages.

[Their docs](https://github.com/streamich/react-use/blob/master/docs/Usage.md) say that we can use individual imports, which should include only the minimal required bundle (although regular named import should do the same with tree-shaking).

So we may be fine with just including the library as a direct dependency, but ideally we should use a self-written hook instead, to make the dependency list as short as possible.

kudlajz: I'm wondering if it's even worth it at this point 😄  maybe we can just keep it duplicated in each theme..

kudlajz: Yeah, I'll change it there.

PR #20, opened 2022-04-12
[TF-665] Extract SEO components from Bea Theme

riffbyte:
I was thinking on merging the `StorySeo` and `PageSeo` components together, but decided that the component would become too overcrowded. On the other hand, it would simplify the usage a lot, requiring only one component to be included in the theme. 
I'm also not entirely sure on how the prop override should work - should it completely override our default props, or add on top of them?

Curious what you think @kudlajz :)


kudlajz (approved):
I think it's fine like this. Combining the two components doesn't seem like it's worth it.

PR #21, opened 2022-04-12
[TF-667] Extract Sitemap building code from Bea Theme

oleg-semyonov:


src/components-nextjs/Sitemap/SitemapBuilder.ts
```diff
@@ -0,0 +1,56 @@
...
+export class SitemapBuilder {
+    private baseUrl: string;
+
+    private urls: SitemapUrl[] = [];
+
+    constructor(baseUrl: string) {
+        this.baseUrl = baseUrl;
+    }
+
+    public addUrl(loc: string) {
```
kudlajz, line 12: Would be nice to use a full name for this, which is a `location` I assume? This applies to all occurences in this file.

src/components-nextjs/Sitemap/types.ts
```diff
@@ -0,0 +1,5 @@
+export type SitemapUrl = {
+    loc: string;
+    changefreq?: string;
+    priority?: string;
+};
```
kudlajz, line 5: It would be nice to use more descriptive property names like `location` and `changeFrequency`.

src/components-nextjs/Sitemap/createPaths.ts
```diff
@@ -0,0 +1,21 @@
...
+    const storiesUrls = stories.map(({ slug }) => `/${slug}`);
+    const categoriesUrls = categories
+        .map((category) => {
+            const translations = Object.values(category.i18n);
+            const allSlugs = translations
+                .map(({ slug }) => slug || '')
+                .filter(Boolean)
+                .reduce(
+                    (slugs, slug) => (slugs.includes(slug) ? [...slugs] : [...slugs, slug]),
+                    [] as string[],
```
kudlajz, line 13: I think you can do `.reduce<string[]>(` instead of the typecast.

kudlajz (approved):

src/components-nextjs/Sitemap/SitemapBuilder.ts
oleg-semyonov: Right, will do!

src/components-nextjs/Sitemap/createPaths.ts
oleg-semyonov: Agree, will check!

PR #25, opened 2022-04-26
Improve CI performance and fix node version

aspirisen:
Discussion https://prezly.slack.com/archives/C029LTE0M/p1651013990025259


package.json
```diff
@@ -29,8 +29,8 @@
   "homepage": "https://github.com/prezly/theme-kit-nextjs#readme",
   "scripts": {
     "clean": "rimraf build/ node_modules/",
-    "prebuild": "npm run clean && npm install",
```
aspirisen, line 32: There is an issue, because it runs npm install before build and override all the linked packages, and because of that React complains about several versions

package.json
```diff
@@ -62,8 +62,11 @@
     "@types/react": "^17.0.38",
     "@types/react-dom": "^17.0.11",
     "eslint": "^8.8.0",
+    "next": "^12.x",
```
aspirisen, line 65: There is one thing that i do not clear understand, but looks like PNPM doesn't install peer dependencies

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Build
```
oleg-semyonov, line 1: Is this action intended to only check that build can go through?

package.json
oleg-semyonov: You need to move these commands somewhere else (e.g. in the `prerelease` script), cause we did have issues with `build` and `node_modules` directories not being cleaned before building a release.

package.json
oleg-semyonov: NPM ditched peer deps install since version 3, but then reintroduced it in version 7, I guess PNPM just didn't catch up :)

oleg-semyonov (changes requested):
I'm not sure about the `prebuild` script, please see my comment.

oleg-semyonov (approved):

PR #31, opened 2022-05-23
[TF-750] Move prop fetching logic for pages into the Theme Kit

oleg-semyonov:
TODO: Add documentation on the helper methods.

The code is basically taken from the Bea theme, with some wrapping logic to allow for passing custom options.

### DIFF
--- a/src/adapter-nextjs/page-props/category.ts
+++ b/src/adapter-nextjs/page-props/category.ts
@@ -1,0 +1,74 @@
+import type { ExtraStoryFields, Story } from '@prezly/sdk';
+import type { GetServerSidePropsContext, GetServerSidePropsResult } from 'next';
+
+import { DEFAULT_PAGE_SIZE } from '../../data-fetching';
+import type { PaginationProps } from '../../infinite-loading/types';
+import { getNewsroomServerSideProps } from '../getNewsroomServerSideProps';
+import { processRequest } from '../processRequest';
+
+import type { PropsFunction } from './lib/types';
+
+export interface CategoryPageProps<StoryType extends Story = Story> {
+    stories: StoryType[];
+    pagination: PaginationProps;
+}
+
+interface Options {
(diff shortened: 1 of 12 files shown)


src/adapter-nextjs/page-props/category.ts
```diff
@@ -0,0 +1,74 @@
...
+                // TODO: This is temporary until return types from API are figured out
+                stories: stories as StoryType[],
+                pagination: {
+                    itemsTotal: storiesTotal,
+                    currentPage: page ?? 1,
+                    pageSize,
+                },
+                ...(typeof customProps === 'function'
+                    ? await (customProps as PropsFunction<CustomProps>)(context, serverSideProps)
+                    : customProps),
```
oleg-semyonov, line 69: In some cases the custom page props depend on the request context or `newsroomContextProps`, this allows to pass a function that derives custom props from these.

e1himself (commented):
I've reviewed this PR. The code looks good.

However, I don't have much of an idea of what's going on here. Totally unfamiliar with the rest of the codebase.

So it would be great to get a review from one of the "maintainers" :)

src/adapter-nextjs/page-props/category.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import type { PropsFunction } from './lib/types';
+
+export interface CategoryPageProps<StoryType extends Story = Story> {
+    stories: StoryType[];
+    pagination: PaginationProps;
+}
+
+interface Options {
+    storiesIncludeFields?: (keyof ExtraStoryFields)[];
```
kudlajz, line 17: Why not just call this the same as the interface - `extraStoryFields`? Seems more readable.

kudlajz (approved):

src/adapter-nextjs/page-props/category.ts
oleg-semyonov: Yeah, that makes much more sense! Thank you!

PR #66, opened 2022-07-01
[TF-930] Limit available story translations based on story's visibility

oleg-semyonov:


src/newsroom-context/hooks/useGetTranslationUrl.ts
```diff
@@ -8,6 +8,32 @@ import { getCategoryHasTranslation, getCategoryUrl } from '../../utils';
 import { useCurrentCategory } from './useCurrentCategory';
 import { useCurrentStory } from './useCurrentStory';
 
+// Pulled from SDK types to not leak the `@prezly/sdk` dependency into the client bundle
+enum StoryVisibility {
+    PUBLIC = 'public',
+    EMBARGO = 'embargo',
+    PRIVATE = 'private',
+    CONFIDENTIAL = 'confidential',
+}
```
oleg-semyonov, line 17: Had to pull this enum in, so that `@prezly/sdk` won't leak into the client bundle.
@e1himself do you know any alternatives for such situations? I already did that in other places and it's not super optimal, but I can't seem to find a better solution.

src/newsroom-context/hooks/useGetTranslationUrl.ts
```diff
@@ -8,6 +8,32 @@ import { getCategoryHasTranslation, getCategoryUrl } from '../../utils';
...
+
+    if (visibility === StoryVisibility.PRIVATE) {
+        return [StoryVisibility.PRIVATE, StoryVisibility.PUBLIC];
+    }
+
+    if (visibility === StoryVisibility.CONFIDENTIAL) {
+        return [StoryVisibility.CONFIDENTIAL, StoryVisibility.PRIVATE, StoryVisibility.PUBLIC];
+    }
+
+    return [StoryVisibility.PUBLIC];
```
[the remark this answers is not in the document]
oleg-semyonov: Well duh, I don't know why, but something inside of me told "don't use the switch" this one time 🤔 
I'll update it to switch, it would be much cleaner.

src/newsroom-context/hooks/useGetTranslationUrl.ts
oleg-semyonov: Currently Next.js seems to not do it. Maybe it's only enabled when the app itself is compiled as a module, but that currently doesn't work well on Next.js (although they claim to support it out of the box)

e1himself: > Currently Next.js seems to not do it. Maybe it's only enabled when the app itself is compiled as a module, but that currently doesn't work well on Next.js (although they claim to support it out of the box)

Ah, I've seen this before. I believe the modern bundlers only do tree-shaking for ESM code. But because our Javascript SDK is compiled to CommonJS only, it's not being performed out-of-the box.

I think that's a good reason to introduce a dual ESM/CommonJS build for the Javascript SDK. Or just go with ESM-only releases.

oleg-semyonov: Yeah, I've been thinking that the time has probably come to move all of our libraries to ESM (although it still feels shaky with some big libraries not really embracing it yet)

PR #98, opened 2022-08-17
[QRF-8] Apply effects on `getAssetsUrl` function

inf1nite-lo0p:
@oleg-semyonov I'll add notes on how to use [yalc](https://github.com/wclr/yalc) for parallel development of packages and themes in a separate PR since that can be applied to our other libraries.

### DIFF
--- a/src/utils/getAssetsUrl.test.ts
+++ b/src/utils/getAssetsUrl.test.ts
@@ -1,31 +1,9 @@
-import type { UploadedImage } from '@prezly/sdk';
 import { ASSETS_URL } from './constants';
 import { getAssetsUrl } from './getAssetsUrl';
 
-const uploadcareImage: UploadedImage = {
-    version: 2,
-    uuid: '43258238-9e71-49fa-a35e-dca5baf16a3b',
-    filename: 'prezly-photographe.png',
-    mime_type: 'image/png',
-    size: 74234,
-    original_width: 1162,
-    original_height: 369,
-    effects: ['/crop/369x369/0,0/', '/preview/'],
-};
-
 describe('getAssetsUrl', () => {
-    it('generates correct asset URL without effects', () => {
-        const image = { ...uploadcareImage, effects: [] };
-        expect(getAssetsUrl(image)).toBe(`${ASSETS_URL}/${image.uuid}/`);
-    });
-
-    it('generates correct asset URL with effects', () => {
-        expect(getAssetsUrl(uploadcareImage)).toBe(
-            `${ASSETS_URL}/${uploadcareImage.uuid}/-${uploadcareImage.effects[0]}-${uploadcareImage.effects[1]}`,
-        );
-    });
-
     it('generates correct asset URL with uuid passed to it', () => {
-        expect(getAssetsUrl(uploadcareImage.uuid)).toBe(`${ASSETS_URL}/${uploadcareImage.uuid}/`);
+        const uuid = 'foo-bar';
+        expect(getAssetsUrl(uuid)).toBe(`${ASSETS_URL}/${uuid}/`);
     });
 });
--- a/src/utils/getAssetsUrl.ts
+++ b/src/utils/getAssetsUrl.ts
@@ -1,17 +1,9 @@
-import type { UploadedImage } from '@prezly/sdk';
-
 import { ASSETS_URL } from './constants';
 
 /**
- * @param image The `UploadedImage` object or uuid of the asset hosted on Uploadcare
+ * @param uuid UUID of the assets hosted on Uploadcare
  * @returns Full asset URL that can be used as a source for <img>, <video> and <a> tags.
  */
-export function getAssetsUrl(image: UploadedImage | string) {
-    if (typeof image === 'string') {
-        return `${ASSETS_URL}/${image}/`;
-    }
-    if (image.effects.length > 0) {
-        return `${ASSETS_URL}/${image.uuid}/-${image.effects.join('-')}`;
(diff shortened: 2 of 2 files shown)


src/utils/getAssetsUrl.test.ts
```diff
@@ -1,10 +1,36 @@
 import { ASSETS_URL } from './constants';
 import { getAssetsUrl } from './getAssetsUrl';
 
+const uploadCareImage = {
```
oleg-semyonov, line 4: Uploadcare is written as a single word, let's not mess up their brand name :)
Please update to `uploadcare` or `Uploadcare` across your code.

oleg-semyonov, line 4: Also, just to be safe, it's better to specify the type here, so that we're certain that the object structure is correct and up-to-date.

src/utils/getAssetsUrl.test.ts
```diff
@@ -1,10 +1,36 @@
...
+    download_url:
+        'https://cdn.uc.assets.prezly.com/43258238-9e71-49fa-a35e-dca5baf16a3b/-/crop/369x369/0,0/-/preview/-/inline/no/prezly-photographe.png',
+    cdnUrl: 'https://cdn.uc.assets.prezly.com/43258238-9e71-49fa-a35e-dca5baf16a3b/-/crop/369x369/0,0/-/preview/',
+};
+
 describe('getAssetsUrl', () => {
-    it('generates correct asset URL', () => {
-        const uuid = 'test-uuid';
+    it('generates correct asset URL without effects', () => {
+        const image = { ...uploadCareImage, effects: [] };
```
oleg-semyonov, line 21: This is the only place where we actually need to copy the object, the other two cases can use the object directly, since it's not manipulated in any way.

src/utils/getAssetsUrl.test.ts
```diff
@@ -1,10 +1,36 @@
...
 
+const uploadCareImage = {
+    version: 2,
+    uuid: '43258238-9e71-49fa-a35e-dca5baf16a3b',
+    filename: 'prezly-photographe.png',
+    mime_type: 'image/png',
+    size: 74234,
+    original_width: 1162,
+    original_height: 369,
+    effects: ['/crop/369x369/0,0/', '/preview/'],
```
oleg-semyonov, line 12: Not sure if `preview` effect is present on real images, but for our use-case it's better to remove it when applying effects to the URL. Can you update the logic to filter out, as well as cover that in the tests?

oleg-semyonov (changes requested):
Nice work! Really glad to see tests updated :)
I have a couple remarks on how to make it a bit cleaner and cover some extra cases.

src/utils/getAssetsUrl.ts
```diff
@@ -1,9 +1,17 @@
...
  */
-export function getAssetsUrl(uuid: string) {
-    return `${ASSETS_URL}/${uuid}/`;
+export function getAssetsUrl(image: UploadedImage | string) {
+    if (typeof image === 'string') {
+        return `${ASSETS_URL}/${image}/`;
+    }
+    if (image.effects.length > 0) {
+        return `${ASSETS_URL}/${image.uuid}/-${image.effects.join('-')}`;
+    }
```
e1himself, line 15: I wish we had this functionality implemented in a single place and reused everywhere.

There is `prezly/uploadcare` package that has it: https://github.com/prezly/uploadcare/

But I guess there was a reason to not use it. Is this to reduce the bundle size @oleg-semyonov ?

e1himself:
> TS says these props don't exist in UploadedImage type, but they actually do (Should fix the types in the @prezly/sdk package)

They are deprecated and should not be used. Therefore they are not in the TS definitions.

src/utils/getAssetsUrl.ts
oleg-semyonov: Yes, that was purely bundle size decision. But I agree that it's better to have everything in one place. Maybe if the bundle size impact is not that big, we can include that it in the deps.

oleg-semyonov: @MohammadxAli can you check the impact on the bundle size? It would make more sense to use the `uploadcare` package as Ivan suggested.

e1himself:
> Aside from that, I have checked the [bundle size of @prezly/uploadcare](https://bundlephobia.com/package/@prezly/uploadcare@2.3.4), which is 12.7KB minified, which is huge IMO for our simple use cases.

@MohammadxAli but did you check bundle size impact after you include the `UploadcareImage` from `@prezly/uploadcare`? I believe that after tree-shaking the situation can be much more optimistic than +12KB of minified JS.

oleg-semyonov:
Hey @MohammadxAli, any updates on this one? 🤔

oleg-semyonov:
Well it ended up being exactly 12 Kb difference 🤷🏻‍♂️ 
The overall increase is actually not that much (5%), so I would say that having a single source-of-truth for Uploadcare-related stuff justifies it.
We have much more work to do in other areas, as well as porting our libraries to ES modules to allow for proper tree-shaking.

oleg-semyonov:
> I guess the problem is with conflicting react versions, since theme-kit uses react 18 and themes using react 17.

For testing purposes, you can either temp-install React 18 on your theme, or temporary roll back to React 17 on Theme Kit.

oleg-semyonov (approved):
The fix looks good!

We'll need to update the `getAssetsUrl` method either way, but let's do it in a separate PR.

PR #106, opened 2022-08-26
[QRF-111] Add a contributing guide

MohammadxAli:
I realized writing docs is much harder than coding :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ All work on @prezly/theme-kit-nextjs happens directly on GitHub. Both core team 
 
 Assuming you have cloned the repo and installed dependencies on your machine, the very next step is to develop the package.
 
-Since theme-kit provides a collection of base functions, hooks, and utilities that works with [Prezly themes](https://github.com/prezly?q=theme-nextjs&type=all&language=&sort=) under the hood, in order to make changes and test things out on your side, you're gonna need to keep a theme running in parallel with the theme-kit.
+Since theme-kit provides a collection of base functions, hooks, and utilities that work with [Prezly themes](https://github.com/prezly?q=theme-nextjs&type=all&language=&sort=) under the hood, in order to make changes and test things out on your side, you're gonna need to keep a theme running in parallel with the theme-kit.
 
 So if you haven't already, go ahead and follow the [instructions on Bea theme](https://github.com/prezly/theme-nextjs-bea#quick-start) to get it up and running.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,6 +3,8 @@
 First off, thanks for taking the time to contribute! 🎉  
 All work on @prezly/theme-kit-nextjs happens directly on GitHub. Both core team members from Prezly and external contributors send pull requests which go through the same review process.
 
+---
+
 ## Development
 
 Assuming you have cloned the repo and installed dependencies on your machine, the very next step is to develop the package.
@@ -12,6 +14,24 @@ Since theme-kit provides a collection of base functions, hooks, and utilities th
 So if you haven't already, go ahead and follow the [instructions on Bea theme](https://github.com/prezly/theme-nextjs-bea#quick-start) to get it up and running.
 
 As you can already see in the theme's `package.json` file, we're using `@prezly/theme-kit-nextjs` package. So we need to make sure that our theme is using our local version of theme-kit and not the one's in the theme's `node_modules` folder. How do we achieve that?
+
+### Option 1 - Using `npm link`
+
+In the **theme-kit's project root** folder run the following command:
+
+```
+npm link
+```
+
+Then go back to **theme's project root** and run:
+
+```
+npm link @prezly/theme-kit-nextjs
+```
+
+This will create a symbolic link between local version of the package and theme.
+
+### Option 2 - Using `yalc`
 
 There's this awesome package called [yalc](https://github.com/wclr/yalc), which serves the purpose for our use case! So let's install it:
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
+# Contributing to @prezly/theme-kit-nextjs
+
+First off, thanks for taking the time to contribute! 🎉
+
+Assuming you have cloned the repo and installed dependencies on your machine, the very next step is to develop the package.
+
+Since theme-kit provides a collection of base functions, hooks, and utilities that works with [Prezly themes](https://github.com/prezly?q=theme-nextjs&type=all&language=&sort=) under the hood, in order to make changes and test things out on your side, you're gonna need to keep a theme running in parallel with the theme-kit.
```
oleg-semyonov, line 7:
```suggestion
Since theme-kit provides a collection of base functions, hooks, and utilities that work with [Prezly themes](https://github.com/prezly?q=theme-nextjs&type=all&language=&sort=) under the hood, in order to make changes and test things out on your side, you're gonna need to keep a theme running in parallel with the theme-kit.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+```
+yalc publish
+```
+
+This will publish the package in the yalc's local store, then go back to your **theme's project root**, and run the following commands:
+
+```bash
+npm uninstall -S @prezly/theme-kit-nextjs
+yalc add @prezly/theme-kit-nextjs
+npm install
```
oleg-semyonov, line 38: AFAIK There is no need to run `uninstall`. I got it to work with just the `yalc` command.

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+What did we do? We simply removed the theme's `@prezly/theme-kit-nextjs` package in `node_modules` and told our theme to use the local version of the package on our file system!
+
+Now the fun part begins, go ahead and hack the package (Make changes) :) Once you're done, simply run the following in **theme-kit's project root** folder:
+
+```bash
+yalc push
+```
```
oleg-semyonov, line 47: It's better to run `yalc publish --push`, that will make all of the projects that have the YALC dep installed automatically update it. See https://github.com/wclr/yalc#pushing-updates-automatically-to-all-installations

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+
+It acts like you have released a new version of your package that your theme can use.
+And don't forget to run `npm install` in the theme's folder _if you have changed dependencies_ for the **theme-kit**.
+
+If the changes are not automatically picked up by theme's dev server, restart the **theme's dev server**.
```
oleg-semyonov, line 52: I found that I needed to delete the `.next` cache folder each time I've updated the dependency, did you need to do that as well? Somehow I couldn't make the changes to be updated live (although it was working that way when I used `npm link`)

oleg-semyonov (commented):
Great start, Ali! 👍🏻 

I have some questions and remarks on the commands.

Also, the primary purpose of CONTRIBUTING.md file is to outline various ways to contribute to the repo (as well as explaining rules and checks that particular repo has in place), not just how to setup the dev environment.

I suggest you take a look at these examples in big repos for inspiration:
https://github.com/github/docs/blob/main/CONTRIBUTING.md
https://github.com/angular/angular/blob/main/CONTRIBUTING.md
https://reactjs.org/docs/how-to-contribute.html
https://github.com/facebook/jest/blob/main/CONTRIBUTING.md

There's also [this nice article by Mozilla](https://mozillascience.github.io/working-open-workshop/contributing/) that can help structure the document.

Not expecting you to write the best version of the file from the get go, but I would like to see a bit more structure to it before I can hit "Approve" :)

I'll also request additional reviews so that we can have more opinions on the topic.

CONTRIBUTING.md
MohammadxAli: Hmm, I guess for some reason it didn't work for me without uninstalling it first 🤔

CONTRIBUTING.md
MohammadxAli: Yeah, good point! Had the same issue, I deleted `node_modules` and `.next` and I guess deleting `.next` fixed that for me, I guess it makes sense to add `npm link` trick as a secondary option too.

There's also a `yalc link` command which I haven't tested it out yet, I'll try it and make sure to add it if it makes dev easier.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+npm link
+```
+
+Then go back to **theme's project root** and run:
+
+```
+npm link @prezly/theme-kit-nextjs
+```
+
+This will create a symbolic link between local version of the package and theme.
```
e1himself, line 32: Does `npm link` option even work? Because every time I've tried using it for testing a theme app with an updated content-renderer it never worked -- `npm` installs to versions of React breaking the build.

With `yalc` I had to struggle too, but eventually it worked.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+    ```
+
+2. Changes are not picked up by yalc after changing deps:  
+   Try deleting `.next` folder in theme and restart the dev server.
+
+---
+
+## Bugs
+
+We use [Linear](https://linear.app) to track issues for our bugs at Prezly, but we also keep an eye on [GitHub issues](https://github.com/prezly/theme-kit-nextjs/issues) for bugs reported from the community. Before filing a new issue, try to make sure your problem doesn't already exist.
```
e1himself, line 107: Does it make sense to mention Linear? I mean, what does it change for an open-source contributor? GitHub issues is still the only option really.

e1himself (approved):
Looks good :+1: 

Some minor comments below.

CONTRIBUTING.md
MohammadxAli: @e1himself For me it used to work at first, then for some unknown reason, it stopped working :) So I started using `yalc`. 

It's always nice to have a secondary option and it has less steps to set up I would say let's have it in place, however, I guess we can move it to 2nd option.

I have mentioned two different versions of react fix in the Common Errors section, I'll add a link to that.

PR #172, opened 2022-12-08
[QRF-593] Fix - Improper Stories API calls

e1himself:


aspirisen (approved):

oleg-semyonov (approved):
Thanks a lot Ivan! ❤️

When Dmitrii was updating the SDK, these lines remained unchanged, so didn't even see them in the diff, and I completely forgot that this property needed updating.
Glad that you figured it out so quickly!

e1himself:
> When Dmitrii was updating the SDK, these lines remained unchanged, so didn't even see them in the diff, and I completely forgot that this property needed updating.
> Glad that you figured it out so quickly!

Yeah, it was nearly impossible to spot it, as anyone would expect Typescript to catch this.

PR #178, opened 2022-12-16
[QRF-642] Only generate alternate links for published stories

kudlajz:


src/newsroom-context/hooks/useGetTranslationUrl.ts
```diff
@@ -63,8 +64,10 @@ function getTranslationUrl(
         const allowedVisibilityValues = getAllowedTranslationVisibilityValues(currentStory);
 
         const translatedStory = currentStory.translations.find(
-            ({ culture, visibility }) =>
-                culture.locale === localeCode && allowedVisibilityValues.includes(visibility),
+            ({ culture, lifecycle_status, visibility }) =>
+                culture.locale === localeCode &&
+                Story.isPublished(lifecycle_status) &&
+                allowedVisibilityValues.includes(visibility),
```
kudlajz, line 70: I was wondering if we even need `getAllowedTranslationVisibilityValues` because we don't want to generate alternate links for private or embargo stories anyway? 🤔 As those are supposed to be "not public", right?

oleg-semyonov: I can't remember the exact ticket, but we had a discussion, where we agreed on this logic, most likely 'cause it was a requirement for a customer. It was definitely done on purpose, so there should be some reason behind it :)

oleg-semyonov (approved):

PR #183, opened 2022-12-23
[QRF-673] Add `withHighlightedStory` parameter to properly handle highlighted stories

oleg-semyonov:
The issue that this PR is intended to fix was present on most of our themes from day one, but somehow it was either overlooked or ignored.

For themes having a dedicated "highlighted story" card (either to display latest or pinned story), the card grid does not look even, since the grid for other stories is missing one item that is moved to "highlighted".

With current abstractions, it's quite hard to have separate page sizes for the initial load and subsequent loads performed by infinite loading.

This PR adds `withHighlightedStory` to `getHomepageServerSideProps ` as well as the API methods that it uses. 
It will still use the `pageSize` parameter for infinite loading, but will load one more story for the first page, and also offset all subsequent page loads by 1 to prevent duplicate cards.

Currently this PR is blocked by API bug -> [QRF-699](https://linear.app/prezly/issue/QRF-699/sdkstoriessearch-api-endpoint-treats-offset-parameter-incorrectly)

### DIFF
--- a/src/data-fetching/api/PrezlyApi.ts
+++ b/src/data-fetching/api/PrezlyApi.ts
@@ -159,6 +159,8 @@ export class PrezlyApi {
             return useHighlightedStory && baseOffset ? baseOffset + 1 : baseOffset;
         }
 
+        console.log({ limit, offset: getPageOffset(), useHighlightedStory });
+
         const { stories, pagination } = await this.searchStories({
             limit,
             offset: getPageOffset(),
--- a/src/data-fetching/api/PrezlyApi.ts
+++ b/src/data-fetching/api/PrezlyApi.ts
@@ -43,6 +43,11 @@ interface GetStoriesOptions {
     pinning?: boolean;
     include?: (keyof Story.ExtraFields)[];
     localeCode?: string;
+    /**
+     * When set to `true`, the result for the first page will include one extra story to place as highlighted story.
+     * This will offset each subsequent page by 1 story to account for that.
+     */
+    useHighlightedStory?: boolean;
 }
 
 interface GetGalleriesOptions {
@@ -138,13 +143,25 @@ export class PrezlyApi {
         pinning = false,
         include,
         localeCode,
+        useHighlightedStory,
     }: GetStoriesOptions = {}) {
         const sortOrder = getChronologicalSortOrder(order, pinning);
         const query = JSON.stringify(getStoriesQuery(this.newsroomUuid, undefined, localeCode));
 
+        const limit = useHighlightedStory && (!page || page === 1) ? pageSize + 1 : pageSize;
+
+        function getPageOffset() {
+            if (typeof page === 'undefined') {
+                return undefined;
+            }
+
+            const baseOffset = (page - 1) * pageSize;
+            return useHighlightedStory && baseOffset ? baseOffset + 1 : baseOffset;
+        }
+
         const { stories, pagination } = await this.searchStories({
-            limit: pageSize,
-            offset: typeof page === 'undefined' ? undefined : (page - 1) * pageSize,
+            limit,
+            offset: getPageOffset(),
             sortOrder,
             query,
             include,
@@ -163,7 +180,7 @@ export class PrezlyApi {
             order = DEFAULT_SORT_ORDER,
             include,
             localeCode,
-        }: Omit<GetStoriesOptions, 'pinning'> = {},
+        }: Omit<GetStoriesOptions, 'pinning' | 'useHighlightedStory'> = {},
     ) {
         const sortOrder = getChronologicalSortOrder(order);
         const query = JSON.stringify(getStoriesQuery(this.newsroomUuid, category.id, localeCode));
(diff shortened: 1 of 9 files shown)


src/infinite-loading/useInfiniteStoriesLoading.ts
```diff
@@ -10,9 +10,11 @@ import { useInfiniteLoading } from './useInfiniteLoading';
 async function fetchStories<T extends Story = Story>(
     page: number,
     pageSize: number,
+    useHighlightedStory: boolean,
     category?: Category,
     locale?: LocaleObject,
     include?: (keyof Story.ExtraFields)[],
+    pinning?: boolean,
```
oleg-semyonov, line 17: @e1himself I didn't notice that when reviewing your PR, but `pinning` was not taken into account for subsequent infinite loading of stories. I fixed it in this PR, as it was related.

src/infinite-loading/useInfiniteLoading.ts
```diff
@@ -26,8 +26,9 @@ export function useInfiniteLoading<T>({
     const [data, setData] = useState<T[]>(initialData);
     const [currentPage, setCurrentPage] = useState(pagination.currentPage);
 
-    const { itemsTotal, pageSize } = pagination;
-    const pagesTotal = Math.ceil(itemsTotal / pageSize);
+    const { itemsTotal, pageSize, useHighlightedStory } = pagination;
+    const adjustedItemsTotal = useHighlightedStory ? itemsTotal : itemsTotal - 1;
+    const pagesTotal = Math.ceil(adjustedItemsTotal / pageSize);
```
oleg-semyonov, line 31: With `useHighlightedStory`, the first page will be one story larger than the others. This logic accounts for that.

src/data-fetching/api/PrezlyApi.ts
```diff
@@ -138,13 +143,18 @@ export class PrezlyApi {
...
         include,
         localeCode,
+        useHighlightedStory,
     }: GetStoriesOptions = {}) {
         const sortOrder = getChronologicalSortOrder(order, pinning);
         const query = JSON.stringify(getStoriesQuery(this.newsroomUuid, undefined, localeCode));
 
+        const limit = useHighlightedStory && (!page || page === 1) ? pageSize + 1 : pageSize;
+        const baseOffset = typeof page !== 'undefined' ? (page - 1) * pageSize : undefined;
+        const offset = useHighlightedStory && baseOffset ? baseOffset + 1 : baseOffset;
```
oleg-semyonov, line 153: There is currently a bug in the API, which does not allow to use `offset` that is not equally dividable by the `limit`. This is not how these parameters should work, so until this bug is resolved, this PR can't be merged.
See [QRF-699](https://linear.app/prezly/issue/QRF-699/sdkstoriessearch-api-endpoint-treats-offset-parameter-incorrectly)

e1himself: Let me fix the backend now. Then we don't need these hacks. 

Agree?

oleg-semyonov: This code is already made with API working as intended in mind. If we use the highlighted story, we need to offset each page by one.
E.g with page size of 10, without highlighted story, the pages would have offsets of 0, 10, 20, etc.
With highlighted story, since the first page is 11 stories, the offsets would be 0, 11, 21, etc

oleg-semyonov: Still, since the offset property does not work as expected, the backend fx still needed

src/data-fetching/api/PrezlyApi.ts
```diff
@@ -43,6 +43,11 @@ interface GetStoriesOptions {
     pinning?: boolean;
     include?: (keyof Story.ExtraFields)[];
     localeCode?: string;
+    /**
+     * When set to `true`, the result for the first page will include one extra story to place as highlighted story.
+     * This will offset each subsequent page by 1 story to account for that.
+     */
+    useHighlightedStory?: boolean;
```
e1himself, line 50: Would it make sense to rename this to `withHighlightedStory`? To match our other behavior modification flags naming, and not make it look like a React hook.

src/data-fetching/api/PrezlyApi.ts
e1himself: Ah, I see. Thanks :+1: 

One suggestion though. This code, if extracted to a function, would look much more readable (IMO):

```ts
function toPaginationParams(params: { page?: number, pageSize: number, withHighlightedStory: boolean }) {
    const { page = 1, pageSize, withHighlightedStory } = params;
    const offset = pageSize * (page - 1);
    const limit = pageSize;

    if (withHighlightedStory && page === 1) {
        return { offset: 0, limit: limit + 1 };
    }

    if (withHighlightedStory && page > 1) {
        return { offset: offset + 1, limit };
    }

    return { offset, limit };
}
```

oleg-semyonov: Yes, that looks much nicer! I'll address this and other comment in a bit.

src/data-fetching/api/PrezlyApi.ts
oleg-semyonov: 7c8c6d80aaee6928ac8d410f107b6c630446ca10

src/data-fetching/api/PrezlyApi.ts
oleg-semyonov: 9d898aeafef3fed808b8011aa55608654fc897f5

e1himself (approved):

e1himself:
I believe this one is safe to merge.

PR #215, opened 2023-02-02
[QRF-950] Improve hrefLang seo

aspirisen:
https://linear.app/prezly/issue/QRF-950/fix-hreflang-for-themes

see https://linear.app/prezly/issue/QRF-951#comment-34960080 for some explanation


src/intl/getFallbackLocale.ts
```diff
@@ -0,0 +1,26 @@
...
+import { LocaleObject } from './localeObject';
+
+/**
+ * Fallback locales for each language.
+ * For example, a story has no en(global) locale, then we can take some fallback locale from here.
+ */
+export const localeFallback: Record<LangCode, LocaleObject[]> = {
+    en: [LocaleObject.fromAnyCode('en-US'), LocaleObject.fromAnyCode('en-GB')],
+    fr: [LocaleObject.fromAnyCode('fr-FR'), LocaleObject.fromAnyCode('fr-CA')],
+};
```
oleg-semyonov, line 11: I don't get why this is needed. You can use `LocaleObject` methods to find a suiting locale by language code (the part before the dash)

src/intl/locale.ts
```diff
@@ -8,11 +8,15 @@ export const DEFAULT_LOCALE = 'en';
 // We use pseudo locale used for localization testing, to reliably determine if we need to fallback to the default newsroom language
 export const DUMMY_DEFAULT_LOCALE = 'qps-ploc';
```
oleg-semyonov, line 9: What are the `__BRAND__` properties in this file for?

_(can't place a comment on the actual line for some reason)_

src/seo/utils/createAlternateLanguageLink.ts
```diff
@@ -0,0 +1,18 @@
+import type { LocaleObject } from '../../intl';
+
+export function createAlternateLanguageLink(
```
oleg-semyonov, line 3: This method seems redundant. You can just put the logic in the `document` component

oleg-semyonov (changes requested):
This PR made me quite confused.

There is a [Document component](https://github.com/prezly/theme-kit-nextjs/blob/main/src/components-nextjs/document.tsx) exported from theme-kit, which has a TODO for providing alternate locales. That would be the best place to put the logic in, so that all themes can instantly utilize it with zero code changes.

Overall I feel like the logic here is too complex, and some explanation on individual parts would help understand it better. I would expect that simply looping over the newsroom languages would allow us to generate proper links. We shouldn't really provide links to locale codes other than what's defined in the newsroom language settings. Please correct me if I'm missing something.

src/intl/locale.ts
aspirisen: It is a workaround to create a nominal type https://github.com/Microsoft/Typescript/issues/202
These two string are not compatible, so i can't assign one to another

src/intl/getFallbackLocale.ts
aspirisen: it is not a locale code map, it is fallback priority, for example if a story doesn't have global `en` translation but have some other `en-*` locales, here is the priority what should be used for region-less `hrefLang` attribute

aspirisen:
> There is a [Document component](https://github.com/prezly/theme-kit-nextjs/blob/main/src/components-nextjs/document.tsx) exported from theme-kit,

Ah, right, forgot about that, am i right that it should be added in PageSeo https://github.com/prezly/theme-kit-nextjs/blob/main/src/components-nextjs/PageSeo/PageSeo.tsx ?

> simply looping over the newsroom languages 

It is no enough, we should also provide `hrefLang` without region and `x-default` links

src/intl/getFallbackLocale.ts
oleg-semyonov: If that's the case, we'd need much more languages covered than just English and French.
It's also worth making the comments and/or the variable naming a bit more clear about its intended usage (determining priority fallback languages)

src/intl/locale.ts
oleg-semyonov: Is that extra type safety really needed, considering you added the `isGlobal` property on the `LocaleObject` class?

oleg-semyonov:
> Ah, right, forgot about that, am i right that it should be added in PageSeo https://github.com/prezly/theme-kit-nextjs/blob/main/src/components-nextjs/PageSeo/PageSeo.tsx ?

Yes, that would be even better. I forgot that we already had this logic there. Then the TODO in the document component can be removed to prevent confusion.

aspirisen:
>Yes, that would be even better. I forgot that we already had this logic there. Then the TODO in the document component can be removed to prevent confusion.

Good 👍

> x-default should be pretty straightforward without extra methods (get default language and convert to hrefLang properties).

I am not sure, if your default language is French and English is as second language i think it is better to use English if it available as `x-default` since it is more widespread.

src/intl/locale.ts
aspirisen: I think yes, because they are similar and I think it is good to check them on type level

src/seo/utils/createAlternateLanguageLink.ts
aspirisen: This function is use in two different places
image

src/intl/locale.ts
oleg-semyonov: I'm still not sure this is needed, they can be distinguished as well with variable naming or JSDoc comments.

src/seo/utils/mapLanguagesToLocales.ts
```diff
@@ -0,0 +1,31 @@
+import type { NewsroomLanguageSettings } from '@prezly/sdk';
+
+import type { LangCode, LocaleCode } from '../../intl';
+import { LocaleObject } from '../../intl';
+
+/**
+ * Group languages by their global language code (e.g. "en" for "en-US" and "en-GB").
+ */
+export function mapLanguagesToLocales(languages: NewsroomLanguageSettings[]) {
+    let defaultLangLocale: LocaleObject | undefined;
```
oleg-semyonov, line 10: There is already `getDefaultLanguage` method, why would you need to duplicate its logic here?
Especially since the method is called `mapLanguagesToLocales`, it's not really suiting for it to also get the default language.

src/seo/getAlternateLanguageLinks.ts
```diff
@@ -0,0 +1,78 @@
...
+    localesByLangCode.forEach((locales, globalLocaleCode) => {
+        let hasGlobalLocale = false;
+
+        locales.forEach((locale) => {
+            alternateLanguageLocales.push(locale);
+
+            if (!hasGlobalLocale && locale.isGlobal) {
+                hasGlobalLocale = true;
+            }
+        });
```
oleg-semyonov, line 29: This can be simplified quite a bit.
```suggestion
        const hasGlobalLocale = locales.some((locale) => locale.isGlobal);
        alternateLanguageLocales.push(...locales.values());
```

oleg-semyonov (commented):
After spending some more time examining the code, I can see better how it's supposed to work (comments also helped). But I'd like to get @e1himself opinion on this before approving.

src/seo/utils/mapLanguagesToLocales.ts
aspirisen: Just not to loop the same array twice

src/seo/getAlternateLanguageLinks.ts
aspirisen: Also I tried to avoid additional array looping, I assume that `locales.values()` is mapping values to array

src/seo/utils/mapLanguagesToLocales.ts
oleg-semyonov: There would not be enough entries in the `languages` array to make this optimization noticeable, so I would vote for simplicity and readability in this case.

src/seo/getAlternateLanguageLinks.ts
oleg-semyonov: Same as in my previous comment, the size of the array makes gains from such optimizations negligible.

oleg-semyonov (approved):
I'm hitting Approve to prevent any further blocking of this task, however I'd like for @e1himself to check this too prior to merging.

I left additional comments in the previous threads.

src/seo/getAlternateLanguageLinks.ts
aspirisen: Fixed https://github.com/prezly/theme-kit-nextjs/pull/215/commits/b3d696fa8d957067f955b6c7be2f018206185bd5

src/seo/utils/mapLanguagesToLocales.ts
aspirisen: Fixed https://github.com/prezly/theme-kit-nextjs/pull/215/commits/ff561de9b1f92e565e8e70d79101def5a19d11d8

src/intl/locale.ts
aspirisen: Changed https://github.com/prezly/theme-kit-nextjs/pull/215/commits/40ef543d010db69a2cad48d1b77481c111e005b7

e1himself:
Reviewing

src/seo/getAlternateLanguageLinks.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    it('should use en-GB locale as x-default even if other locale is set as default', () => {
+        const links = getAlternateLanguageLinks(
+            [lang('en-GB'), lang('fr-FR', true)],
+            getTranslationUrl,
+            createHref,
+        );
+
+        expect(links).toMatchObject([
+            { href: 'http://localhost:3000/en/translationUrl/en-GB', hrefLang: 'en' },
+            { href: 'http://localhost:3000/en/translationUrl/en-GB', hrefLang: 'en-GB' },
```
e1himself, line 67: Is this valid?

I've checked this article on hreflang best practices here: https://ahrefs.com/blog/hreflang-tags/

They list this typical problem No. 3 in their article:

> ## 3. Page referenced for more than one language in hreflang
> 
> This warning triggers when one or more URLs are referenced for more than one language in hreflang annotations. For example:
> 
> ```
> <link rel="alternate" hreflang="en" href="http://example.com/page.html" />
> <link rel="alternate" hreflang="de" href="http://example.com/page.html" />
> ```
> 
> ### Why it’s an issue
> 
> Each piece of content should only serve one language or language-location. Having two or more contradicting references will confuse search engines, and they may end up ignoring both hreflang attributes.
> 
> ### How to fix
> 
> Review the affected pages, then inspect the URLs that reference the page in their hreflang attributes for errors. Remove the incorrect hreflang attribute to leave only one correct attribute per language.

src/seo/getAlternateLanguageLinks.ts
```diff
@@ -0,0 +1,78 @@
...
+import { getFallbackLocale, LocaleObject } from '../intl';
+
+import { createAlternateLanguageLink } from './utils/createAlternateLanguageLink';
+import { getDefaultHrefLang } from './utils/getDefaultHrefLang';
+import { mapLanguagesToLocales } from './utils/mapLanguagesToLocales';
+
+export function getAlternateLanguageLinks(
+    languages: NewsroomLanguageSettings[],
+    getTranslationUrl: (locale: LocaleObject) => string | undefined,
+    createHref: (locale: LocaleObject, translationUrl: string) => string,
```
e1himself, line 13: These two params look too similar. Can we not combine then into one function `generateUrl(contentLocale, hrefLan) => string | undefined` ?

src/seo/getAlternateLanguageLinks.test.ts
aspirisen: There are two completely different locales in the example, I use the same language part.
I tried https://codesandbox.io/s/confident-curran-esr8gk?file=/b.html here https://app.hreflang.org/ and there were no errors, however, there is indeed an error when different languages point to the same page

src/intl/localeObject.ts
```diff
@@ -12,6 +12,12 @@ export class LocaleObject {
      */
     private localeCode: string;
 
+    /**
+     * Whether the locale is global (without region/culture part).
+     * I.E. `en` is global, while `en-US` is not.
+     */
+    public isGlobal: boolean;
```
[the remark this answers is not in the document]
aspirisen: maybe just super explicit - `isRegionIndependent` or `isRegionless` ?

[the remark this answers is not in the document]
e1himself: Yeah! "region-specific" and "region-independent" naming sounds more clear to me :+1:

src/seo/getAlternateLanguageLinks.test.ts
e1himself: > however, there is indeed an error when different languages point to the same page

Do you think it makes sense to build-in protection logic into the function to make sure we do not generate two alternate links with the same `href` for different `hreflang` values?

src/seo/getAlternateLanguageLinks.ts
```diff
@@ -0,0 +1,78 @@
...
+): AlternateLanguageLink[] {
+    const alternateLanguageFallbackGlobalLinks: AlternateLanguageLink[] = [];
+    const alternateLanguageLocales: LocaleObject[] = [];
+    const localesByLangCode = mapLanguagesToLocales(languages);
+
+    const defaultLanguage = languages.find((language) => language.is_default);
+    const defaultLangLocale = defaultLanguage
+        ? LocaleObject.fromAnyCode(defaultLanguage.code)
+        : undefined;
```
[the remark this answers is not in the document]
aspirisen: Tried to simplify logic https://github.com/prezly/theme-kit-nextjs/pull/215/commits/05dbeda3253754db3259bbfe65281436d27ac40f

src/seo/getAlternateLanguageLinks.test.ts
aspirisen: I am not sure how that could happen since we construct links from languages array. There can be a possibility if server will return duplicated languages, but it is already covered by using "Map". I added test for that case.

src/intl/localeObject.ts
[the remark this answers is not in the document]
aspirisen: Used isRegionIndependent https://github.com/prezly/theme-kit-nextjs/pull/215/commits/a2c2c89db86824fbd0596d47f4452c601d3657be

src/seo/getAlternateLanguageLinks.test.ts
e1himself: > but it is already covered by using "Map". I added test for that case.

I don't think I understood this part.

In these test scenarios, I can see that the code can generate multiple links with the same `href`, but different `hrefLag`, and it's not covered by using the Map. And according to the article I've linked above it is an issue.

aspirisen: There is no error if the same href i used for different regions (or without region) of the same language https://github.com/prezly/theme-kit-nextjs/pull/215#discussion_r1104149754 
As i understand the problem will be if the same page is used for different languages like here
```
<link rel="alternate" hreflang="en" href="http://example.com/page.html" />
<link rel="alternate" hreflang="de" href="http://example.com/page.html" />
```
But I do not quite understand how can we get into this situation

aspirisen:
@e1himself could you re-review this pr when you will have time

aspirisen:
Can I merge this PR ?

oleg-semyonov:
> Can I merge this PR ?

It'd be nice to publish a new preview version with latest changes so that it can be properly tested in a theme. This is quite a big and complex change, so it deserves proper testing.

aspirisen:
@oleg-semyonov here is the PR with this theme-kit

aspirisen:
@oleg-semyonov did you have a chance to check the preview env?

oleg-semyonov (commented):
I checked the the preview you pushed, and it looks fine. The logic part seems solid to me.
I have a couple remarks on the code itself.

src/seo/getAlternateLanguageLinks.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        let isDefaultAddedToLinks = false;
+
+        // First we try to find region independent english locale (en)
+        // If the story has just region independent `en` translation we will use it as default hreflang
+        languagesWithLocales.forEach((binds) =>
+            binds.forEach(({ locale }) => {
+                if (locale.isRegionIndependent && locale.toNeutralLanguageCode() === engFallback) {
+                    pushLocaleToLinks(locale, defaultHrefLang);
+                    isDefaultAddedToLinks = true;
```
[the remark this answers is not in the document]
aspirisen: How can I do that if this code is inside `binds.forEach(({ locale }) => {` function and thus `return` will not stop `populateDefault` function

src/seo/getAlternateLanguageLinks.ts
```diff
@@ -0,0 +1,115 @@
...
+            // we will try to find some link that already has `en` fallback
+            const fallbackFromRegionIndependentLocales = hrefLangLinks.find(
+                (alternate) => alternate.hrefLang === engFallback,
+            );
+
+            if (fallbackFromRegionIndependentLocales) {
+                hrefLangLinks.push({
+                    ...fallbackFromRegionIndependentLocales,
+                    hrefLang: defaultHrefLang,
+                });
```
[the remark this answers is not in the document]
aspirisen: No, because it is not a locale

oleg-semyonov (approved):

oleg-semyonov:
Let's merge it! 👍🏻 
Please be sure to follow SemVer when creating a release, and fill in the Release on GitHub once the package is published.

PR #227, opened 2023-02-24
Add code to ignore junk Sentry error events

oleg-semyonov:
We had lots of junk Sentry errors coming from our themes, mostly caused by misbehaving extensions or people mocking about in the DevTools. This config update should address that and clean up our Sentry dashboard big time 🤞🏻 

Here's only a small subset of errors this is intended to clean up. Note that they are all caused by referring to variables from frameworks/libraries, that are not related to our themes in any way.
image

### DIFF
--- a/src/sentry/common.ts
+++ b/src/sentry/common.ts
@@ -1,5 +1,12 @@
 import type { init } from '@sentry/nextjs';
 
+const IGNORED_EVENT_CULPRITS = [
+    // This usually means that something happened outside of application code.
+    // All instances of errors with this origin that I found usually were caused by referring to variables that couldn't ever exist in our code.
+    // Likely misbehaving browser extensions or people playing around in DevTools trying to break the app.
+    '?(<anonymous>)',
+];
+
 export function getCommonClientOptions(dsn: string, themeName: string): Parameters<typeof init>[0] {
     return {
         dsn,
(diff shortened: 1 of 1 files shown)


src/sentry/common.ts
```diff
@@ -18,6 +25,35 @@ export function getCommonClientOptions(dsn: string, themeName: string): Paramete
             // This error is caused by `beacon.min.js` polyfill in Cloudflare Insights. From what we see, it doesn't affect the user experience.
             'Illegal invocation',
         ],
+
+        // Ignore global code errors that are not related to the application code.
+        // See comments in `IGNORED_EVENT_CULPRITS` const for context on each entry
+        beforeSend: (event) => {
```
kudlajz, line 31: Do you think we could use the [`allowUrls`](https://github.com/prezly/prezly/blob/ac85b8621f5bbfc728972fcd1ea12f5e30968a1c/apps/press/modules/widget/templates/_raven.php#L16-L18) to whitelist our domain only? That way, erros from scripts from other URLs should be ignored.

oleg-semyonov: It would help us, if we had a pre-determined domain, but since themes are multi-tenancy applications, we can't know the domain before we receive the request. And this configuration is executed when the application starts, so we can't get the domain at that stage.

kudlajz: I think you can match partially, so you don't need a full url. Maybe we could have something like `/_next/static/chunks/*`? 🤔 

> As a result, if you add 'foo.com' to it, it will also match on https://bar.com/myfile/foo.com.

See the [documentation](https://docs.sentry.io/platforms/javascript/configuration/options/#allow-urls).

oleg-semyonov: Still not sure, but prolly worth investigating later. My concern is that it won't filter these particular issues, since technically they're happening right on the page, so they would probably pass the URL check.

PR #246, opened 2023-03-23
Install monorepo architecture with Lerna

oleg-semyonov:
This is prep work for splitting the package into a core package and framework adapters.
The setup is similar to [prezly/analytics](https://github.com/prezly/analytics) repo.

The bulk of the diff comes from the package code being moved to `/packages/theme-kit-nextjs` (Folder naming will probably change once the package gets split, for now I didn't want to complicate things too much). You can ignore everything in that folder, except for `tsconfig.json` and `package.json` changes.

Everything in the root folder is set up to manage the packages and run scripts, as well as publish with Lerna.


package.json
```diff
@@ -1,13 +1,8 @@
 {
-  "name": "@prezly/theme-kit-nextjs",
-  "version": "3.2.4",
+  "name": "theme-kit",
+  "version": "0.1.0",
```
oleg-semyonov, line 3: This is a dummy version, the root package is never meant to be published.

packages/theme-kit-nextjs/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "react-dom": "^17.x || ^18.x"
+  },
+  "dependencies": {
+    "@prezly/sdk": "^11.11.0",
+    "@prezly/uploadcare": "^2.3.4",
+    "@react-hookz/web": "^16.0.0",
+    "next-seo": "^5.4.0",
+    "parse-data-url": "^4.0.1"
+  },
+  "devDependencies": {
```
oleg-semyonov, line 47: `peerDependencies` and `dependencies` remained unchanged, the only change here is `devDependencies`, which were stripped of the common deps already declared in the root `package.json` file

oleg-semyonov:
Hey @pestaa @e1himself @kudlajz, would appreciate your review on the new setup :)

pestaa:
Man, I scrolled through this and I applaud the effort. But not familiar with the crucial points around tsconfig, etc. Is it OK that the root is an `esnext` module and the package is `commonjs`?

oleg-semyonov:
> @oleg-semyonov what do you think, maybe this pr is a good place to switch from npm to pnpm ?

I don't think it's really needed at this point, and in any case, I would not mix it up with these changes

kudlajz (approved):
Looks good to me 👍 It's gonna be much easier to work with this as well as adding new adapters I think.

aspirisen (approved):
👍

packages/theme-kit-nextjs/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Prezly
```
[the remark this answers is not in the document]
oleg-semyonov: According to [this StackOverflow guru](https://opensource.stackexchange.com/a/1523), the year in the license resembles the year when the license was applied, so no need to update

PR #275, opened 2023-04-27
[CARE-1382] Make Prezly SDK a peer dependency, narrow types to allow broader SDK compatibility

oleg-semyonov:
**Key changes:**
- `@prezly/sdk` is now a peer dependency. This allows updating the SDK version in consuming applications independently of Theme Kit. Currently any version between v12 and v15 is allowed.
- Added matrix strategy to `typecheck` and `test` workflow to ensure broad SDK version compatibility
- All of the used SDK types were narrowed to the minimum required subset of properties to ensure maximum compatibility with different SDK versions.
- Dropped `isSdkError` helper in favor of `instanceof ApiError` check with the  class exported from `@prezly/sdk`.
- Updated `isEmbargoStory` helper to use `Story.isScheduledEmbargo` check. Bumped SDK to v12 to use the updated `status` check instead of `lifecycle_status`


pestaa (commented):
Seems legit. :+1: But I get a weird sense. You claim this change helps across SDK versions, but I don't see how. Instead of binding to high level types, it's now bound to _specific properties_ of the same high level types. So it's more work with no real upside? Possible I'm missing something.

I will also argue that supporting anything but the latest version is a maintenance trap for small companies like us.

oleg-semyonov:
> Instead of binding to high level types, it's now bound to specific properties of the same high level types. So it's more work with no real upside? Possible I'm missing something.

Here's an example of a scenario that I encountered a while ago:
A Prezly SDK version added a new property to `Story`, that is not used anywhere in the Theme Kit, but is needed in one of the themes. I'd like to be able to just bump the `@prezly/sdk` version on the theme to access the new property, but that causes an error with the Theme Kit, which depends on older version of `@prezly/sdk`, that does not have the necessary property in the types to satisfy the newer type definitions.

Making `@prezly/sdk` a peer dependency would allow updating the theme version independently of the Theme Kit, and will not require us to release a new Theme Kit version on every SDK update (including minor and patch versions).
This approach has some limitations too (we will still have to eventually update the dependency range), but I feel like its giving more flexibility.


> I will also argue that supporting anything but the latest version is a maintenance trap for small companies like us.

I believe I alleviated most of the stress by introducing the version matrix testing. Of course, we'll be bumping the minimum version regularly, but in between those bumps, we'll ensure that any consumer application that hasn't updated the SDK yet will still continue to work.

Also, if we are to follow SemVer, each SDK dependency bump would ideally require a major release of Theme Kit (since it would break any consumer app that has older SDK version in its deps). This is less needed with this approach, since NPM will throw an error about unresolved peer dependency when installing a new Theme Kit version that depends on newer SDK version.


I'm curious what @e1himself thinks about it 🤔

pestaa:
@oleg-semyonov Oh I definitely get the peer dependency part. However I kept reading through the PR and wondered what's the benefit of unpacking types. You're still bound to the same structure but with more effort. I would consider only doing it when the SDK introduces a breaking change, and the Theme Kit needs to connect to both compositions.

oleg-semyonov:
> You're still bound to the same structure but with more effort. I would consider only doing it when the SDK introduces a breaking change, and the Theme Kit needs to connect to both compositions.

Not quiet. With the unpacked types we're only bound to the properties that matter to us, all other properties of the type don't impact us, so if some other properties are added or removed, it won't break the logic or cause a type error.
The only time we'll get an error is if the actual picked properties change, at which point we'd likely want to check and update the code either way.

e1himself (approved):

oleg-semyonov:
I'll go ahead and merge this PR, since it's blocking other tasks for v5 release.
I won't publish the new version until all of the chores are addressed, so if anybody has any objections on this setup, please let me know :)

PR #289, opened 2023-05-10
[CARE-1556] Fix issue with `arr.at is undefined` in safari 14 

fgyimah:
The issue was due to `Array.at()` only being supported in Safari 15.4 and upwards.

Updated to use simple array indexing instead


MohammadxAli (approved):
I guess adding a comment here to explain the reason can be helpful.

Prevents and accidental refactor in the future as well.

PR #291, opened 2023-05-15
[CARE-1613] Exclude private stories in search engines

MohammadxAli:


packages/nextjs/src/components-nextjs/StorySeo/StorySeo.tsx
```diff
@@ -43,8 +43,8 @@ export function StorySeo({ story, noindex }: Props) {
                 title={seoTitle}
                 description={seoDescription}
                 canonical={canonical}
-                noindex={noindex ?? !indexable}
-                nofollow={noindex ?? !indexable}
+                noindex={noindex && !indexable}
+                nofollow={noindex && !indexable}
```
e1himself, line 47: After thinking more, I believe it should be like this:

```tsx
noindex={noindex || !indexable}
nofollow={noindex || !indexable}
```

Same change for `PageSeo.tsx`

e1himself (changes requested):
See my comment above

packages/nextjs/src/components-nextjs/StorySeo/StorySeo.tsx
oleg-semyonov: Was about to post the same comment 😃 
With current code, the seo setting will not take effect if the site is indexable

oleg-semyonov (approved):

packages/nextjs/src/components-nextjs/StorySeo/StorySeo.tsx
e1himself: Inverted boolean flags are tricky! :)

PR #296, opened 2023-05-20
[CARE-1659] Switch from NX to Turborepo

MohammadxAli:


MohammadxAli:
I don't see a real benefit from switching from NX to Turbo yet, I thought it was needed for linking/unlinking a theme to theme-kit, but this can be purely done from a theme side without changing the theme-kit's build system.

PR #297, opened 2023-05-22
[CARE-1572]  Return language by neutral code as 2nd priority 

MohammadxAli:


packages/nextjs/src/intl/languages.ts
```diff
@@ -26,11 +26,29 @@ export function getLanguageFromNextLocaleIsoCode<
...
     if (nextLocaleIsoCode.length >= 2 && nextLocaleIsoCode.length <= 4) {
-        // The order of methods here is reversed from `getShortestLocaleCode`, so that the logic "unwraps" the possible variants with no collisions
-        targetLanguage =
-            getLanguageByExactLocaleCode(languages, locale) ||
-            getLanguageByShortRegionCode(languages, locale) ||
-            getLanguageByNeutralLocaleCode(languages, locale);
+        const languageByExactLocaleCode = getLanguageByExactLocaleCode(languages, locale);
+        const languageByNeutralLocaleCode = getLanguageByNeutralLocaleCode(languages, locale);
+        const languageByShortRegionCode = getLanguageByShortRegionCode(languages, locale);
+
```
MohammadxAli, line 32: I changed the order from:

1. exact
2. short region
3. neutral code

To:

1. exact
2. neutral code
3. short region

However, this change caused some tests to fail. Consequently, I needed to revise the if...else statement in order to ensure that it functions correctly.

oleg-semyonov: As I pointed out in the other PR, we shouldn't change the order of execution, since it's tied to `getShortestLocaleCode` (as was stated in the comment).

I think it's worth it to hop on a pair coding session one more time to figure this out :)

MohammadxAli:
Closing in favor of #301

PR #298, opened 2023-05-22
[CARE-1572] Prioritize languages with same neutral code in `getLanguageByShortRegionCode`

MohammadxAli:
There is a secondary solution to this: #297 

But that being said, I feel like this PR is cleaner and solves the problem in an earlier stage which makes more sense.

### DIFF
--- a/packages/core/src/__mocks__/languages.ts
+++ b/packages/core/src/__mocks__/languages.ts
@@ -1,15 +1,17 @@
 import { Culture, type NewsroomLanguageSettings } from '@prezly/sdk';
 
-/**
- * This data is pulled from The Good Newsroom (with some extra cultures) and reduced to only fields required by tests.
- */
-export const LANGUAGES: Record<
+type MockLanguages = Record<
     string,
     Pick<
         NewsroomLanguageSettings,
         'locale' | 'is_default' | 'public_stories_count' | 'code' | 'company_information'
     >
-> = {
+>;
+
+/**
+ * This data is pulled from The Good Newsroom (with some extra cultures) and reduced to only fields required by tests.
+ */
+export const LANGUAGES: MockLanguages = {
     nl_BE: {
         code: 'nl_BE',
         locale: {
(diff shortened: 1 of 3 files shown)


packages/core/src/__mocks__/languages.ts
```diff
@@ -402,3 +404,88 @@ export const LANGUAGES: Record<
         },
     },
 };
+
+export const LANGUAGES_2: MockLanguages = {
```
oleg-semyonov, line 408: It feels weird having those two objects here. Could we update the tests to work with just a single object?
We can always assemble the languages array manually for each test (not just passing the whole `LANGUAGES`), which would give more control

packages/core/src/intl/languages.ts
```diff
@@ -102,14 +102,26 @@ export function getLanguageByShortRegionCode<
     Language extends Pick<NewsroomLanguageSettings, 'is_default' | 'code' | 'public_stories_count'>,
 >(languages: Language[], locale: LocaleObject): Language | undefined {
     const shortRegionCode = locale.toRegionCode();
+    const neutralLanguageCode = locale.toNeutralLanguageCode();
 
-    // Prefer default language
+    // 1. Prefer language with same neutral language code
```
oleg-semyonov, line 107: This looks incorrect. The method is specifically designed to return a language with matching REGION code part.

To me, the issue is in the order in which this function is executed.
It should be as simple as switching the order of functions in [getLanguageFromNextLocaleIsoCode](https://github.com/prezly/theme-kit-js/blob/c6fff3d81acac45c469248566e9e04625c125157/packages/nextjs/src/intl/languages.ts#L32) function.

However, I can see that the current order of checks in that function corresponds to `getShortestLocaleCode` checks.

All in all, I feel like this requirement of shorting to region code is a huge pain in the butt, and we should get rid of this logic. This used to be a request from older customers, not sure how critical is this nowadays.

oleg-semyonov: As a middle-ground solution, I think you can extract this logic into a separate method and call that one in `getLanguageFromNextLocaleIsoCode` right after exact locale code check.
I'm afraid that hacks like this would make things even worse in the long term

oleg-semyonov (changes requested):
Sorry, can't approve yet. Please see my comments

packages/core/src/intl/languages.ts
MohammadxAli: > This looks incorrect. The method is specifically designed to return a language with matching REGION code part.

Yeah, I also thought about this. But what about the other solution (#297)?

That's exactly what I did there.

MohammadxAli:
Closing in favor of #301

PR #301, opened 2023-05-23
[CARE-1572] Fix region locales incorrect routing

inf1nite-lo0p:
Unfortunately, the `returns correct language with same region` test still fails.

Given the following input as `languages` as an example:

1. `en_DE`
2. `de_DE`

Then `getLanguageFromNextLocaleIsoCode(languages, 'en')` should return `en_DE` but it returns `de_DE`.

But if I apply the same fix from my other PR the test will pass (#302) 👇

https://github.com/prezly/theme-kit-js/blob/dc13d7bb5f73f867e572235210c78ac292ebae39/packages/core/src/intl/languages.ts#L105-L116

PR #302, opened 2023-05-23
[CARE-1572] Priotrizing the language with matching **neutral code & short region** over the only **short region**

MohammadxAli:
Fix for #301 

Prioritize language selection based on a matching combination of neutral code and short region over solely relying on a short region. This is what fixes the issue.

At this point, I'm going crazy with these sequences of PRs.

### DIFF
--- a/packages/core/src/intl/languages.ts
+++ b/packages/core/src/intl/languages.ts
@@ -95,6 +95,17 @@ export function getLanguageByShortRegionCode<
     Language extends Pick<NewsroomLanguageSettings, 'is_default' | 'code' | 'public_stories_count'>,
 >(languages: Language[], locale: LocaleObject): Language | undefined {
     const shortRegionCode = locale.toRegionCode();
+    const neutralLanguageCode = locale.toNeutralLanguageCode();
+
+    const languageWithSameRegionAndNeutralCode = languages.find(({ code }) => {
+        const comparingLocale = LocaleObject.fromAnyCode(code);
+        const hasSameNeutralCode = comparingLocale.toNeutralLanguageCode() === neutralLanguageCode;
+        const hasSameRegionCode = comparingLocale.toRegionCode() === shortRegionCode;
+        return hasSameNeutralCode && hasSameRegionCode;
+    });
+    if (languageWithSameRegionAndNeutralCode) {
+        return languageWithSameRegionAndNeutralCode;
+    }
 
     // Try to look in used cultures first (giving priority to used ones)
     const usedLanguages = getUsedLanguages(languages);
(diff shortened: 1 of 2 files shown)


oleg-semyonov (changes requested):

packages/core/src/intl/languages.ts
```diff
@@ -95,6 +95,17 @@ export function getLanguageByShortRegionCode<
...
+
+    const languageWithSameRegionAndNeutralCode = languages.find(({ code }) => {
+        const comparingLocale = LocaleObject.fromAnyCode(code);
+        const hasSameNeutralCode = comparingLocale.toNeutralLanguageCode() === neutralLanguageCode;
+        const hasSameRegionCode = comparingLocale.toRegionCode() === shortRegionCode;
+        return hasSameNeutralCode && hasSameRegionCode;
+    });
+    if (languageWithSameRegionAndNeutralCode) {
+        return languageWithSameRegionAndNeutralCode;
+    }
```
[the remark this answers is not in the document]
MohammadxAli: You're right :D I just bring back the functionality last night to see if the tests pass! I was so overwhelmed.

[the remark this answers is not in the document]
oleg-semyonov: I feel you brother :D

packages/nextjs/src/intl/languages.ts
```diff
@@ -21,5 +21,19 @@ export function getLanguageFromNextLocaleIsoCode<
...
+        const comparingLocale = LocaleObject.fromAnyCode(code);
+        const hasSameNeutralCode = comparingLocale.toNeutralLanguageCode() === neutralLanguageCode;
+        const hasSameRegionCode = comparingLocale.toRegionCode() === shortRegionCode;
+        return hasSameNeutralCode && hasSameRegionCode;
+    });
+
+    if (languageWithSameRegionAndNeutralCode) {
+        return languageWithSameRegionAndNeutralCode;
+    }
+
```
oleg-semyonov, line 37: I know you're tired of this already, but I was meaning to put this logic inside `getLanguageFromLocaleIsoCode` :D

oleg-semyonov (changes requested):

packages/nextjs/src/intl/languages.ts
MohammadxAli: Ugh, my bad. Sorry :D

Btw, where do you think is the best place for the logic to be? I believe right after this check, right?

https://github.com/prezly/theme-kit-js/blob/01337801b3e5728700b798b02c47e9d7c9486175/packages/core/src/intl/languages.ts#L228

oleg-semyonov: Exactly!

PR #307, opened 2023-05-26
[DEV-10857] Add correct `Cache-Control` headers to every page server request, export `withPrezlyConfig` helper

oleg-semyonov:
This fixes issue with Next 13 overriding the `Cache-Control` headers set in `next.config.js`.


fgyimah (approved):
🥇

PR #308, opened 2023-05-31
[CARE-1683] Exclude `main` branch from running E2E tests in themes

MohammadxAli:
[Source](https://stackoverflow.com/a/58142412/6622233)

### DIFF
--- a/.github/workflows/playwright.shared.yml
+++ b/.github/workflows/playwright.shared.yml
@@ -6,6 +6,11 @@
 
 name: Playwright Tests
 on:
+  push:
+    branches:
+      - '*'
+      # We are explicitly excluding `main` branch from E2E tests: CARE-1683
+      - '!main'
   workflow_call:
     inputs:
       site-name:


fgyimah (approved):

oleg-semyonov (approved):
👍🏻

oleg-semyonov (changes requested):
Sorry, hit approve without realizing that this is an incorrect setup for shared workflow.
Although [GitHub docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) don't mention that other triggers are not allowed, the way the consumer repos are currently set up, is that they would call this workflow either way. So the correct fix is to (sadly) make these `on: push` changes in the theme repos, not in the theme kit. The good news is that you can skip custom themes, since they won't have their site change between `main` branch and preview deployments

MohammadxAli:
Ahhh... That's what I thought when I made the change. Good to know, thanks!

I'll close this and apply the same change in public themes 😄

PR #313, opened 2023-06-06
[DEV-10948] Fetch stories in V4 format, bump SDK version

kudlajz:
- bumped minimum SDK version to 16
- specified V4 format when fetching a single story

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,7 +27,7 @@
     "test": "jest"
   },
   "peerDependencies": {
-    "@prezly/sdk": ">=12 <16"
+    "@prezly/sdk": "16"
   },
   "dependencies": {
     "@prezly/uploadcare": "^2.3.4",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prezly/theme-kit-core",
-  "version": "5.3.0",
+  "version": "5.4.1-alpha.0",
   "description": "Data layer and utility library for developing Prezly themes with JavaScript",
   "main": "build/index.js",
   "types": "build/index.d.ts",
--- a/packages/core/src/data-fetching/api/PrezlyApi.ts
+++ b/packages/core/src/data-fetching/api/PrezlyApi.ts
@@ -228,7 +228,8 @@ export class PrezlyApi {
         );
     }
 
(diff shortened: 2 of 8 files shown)


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@prezly/theme-kit-core",
-  "version": "5.3.0",
+  "version": "5.4.1-alpha.0",
```
kudlajz, line 3: Do I need to revert this before merging or can I merge it with this preview version?

packages/core/src/data-fetching/api/PrezlyApi.ts
```diff
@@ -80,7 +76,9 @@ export class PrezlyApi {
         }
 
         try {
-            const story = await this.sdk.stories.get(uuid);
+            const story = await this.sdk.stories.get(uuid, {
+                formats: [Story.FormatVersion.SLATEJS_V4],
```
kudlajz, line 80: I wasn't sure if I should specify the V4 format in `searchStories` as well? 🤔 Since those are mainly used for listing and such so you're probably not gonna be working with `content`, but I wasn't sure.

Also not sure what the effect on backend would be in this case, when you request a lot of stories and also want them converted to V4 🤔 

Maybe @e1himself or @pestaa can advise here.

e1himself: It's better use the same version for all requests (for completeness). Also, some themes may decide to use `content` on the listing as well (to render `titles` with bold/italic formatting, for example). The performance impact should be negligible, so don't worry about it.

kudlajz: Aight 👍

pestaa: Yep, we do the same in legacy themes. No need to watch out what's going on downstream.

oleg-semyonov (approved):
👍🏻 

Please don't forget to mention breaking changes in the changelog. It's not just the SDK version, but the slate content version v4 could break themes running on v3, right?

oleg-semyonov (approved):
👍🏻 

Please don't forget to mention breaking changes in the changelog. It's not just the SDK version, but the slate content version v4 could break themes running on v3, right?

PR #326, opened 2023-06-15
[CARE-1846] Fix sitemap URLs to support any possible value

MohammadxAli:
It's hard to test it locally, but my guts telling me that `req.headers['x-forwarded-proto']` can be `https,http`


oleg-semyonov (approved):
Great work @MohammadxAli!

PR #336, opened 2023-06-28
[CARE-1898] Implement filtering Galleries by type

oleg-semyonov:


packages/core/src/data-fetching/api/queries.ts
```diff
@@ -58,9 +59,15 @@ export function getChronologicalSortOrder(direction: `${SortOrder.Direction}`, p
     return pinning ? SortOrder.combine(pinnedFirst, chronological) : chronological;
 }
 
-export function getGalleriesQuery() {
+export function getGalleriesQuery(type?: `${NewsroomGallery.Type}`) {
+    const conditions: Object[] = [{ status: { $eq: 'public' } }, { is_empty: { $eq: false } }];
+
+    if (type) {
+        conditions.push({ type });
+    }
```
pestaa, line 67: :ok_hand: 

I believe the `$eq` is optional? So if you want to keep it tidier (in the future), you can perhaps just pass the value you're looking after.

```
{ status: 'public', is_empty: false }
```

But perfect this way too. :)

MohammadxAli (approved):

PR #364, opened 2023-08-11
[DEV-11239] Adds new optional `formats` param to `GetStoryPageServerSideProps`

fgyimah:


packages/nextjs/src/adapter-nextjs/page-props/story.ts
```diff
@@ -16,14 +17,15 @@ import type { PropsFunction } from './lib/types';
 
 export function getStoryPageServerSideProps<CustomProps extends Record<string, any>>(
     customProps: CustomProps | PropsFunction<CustomProps>,
+    formats?: Story.FormatVersion[],
```
e1himself, line 20: Do we need to do the same for `getStoryPageStaticProps()` below and `getStoryPreviewPageServerSideProps()` in the neighbouring storyPreview.ts file?

fgyimah: Yeah agree, making the update

fgyimah: Updated

e1himself (approved):
Looks good :+1:

PR #398, opened 2023-10-18
[DEV-11804] Refactor - Intl subpackage

e1himself:
So this is the changeset on top of git histories merge (see https://github.com/prezly/theme-kit-js/pull/400) to make things work within one monorepo.


nx.json
```diff
@@ -9,14 +9,14 @@
...
-      "outputs": ["{projectRoot}/build", "{projectRoot}/tsconfig.tsbuildinfo"]
+      "dependsOn": ["^build", "prebuild"],
+      "outputs": ["{projectRoot}/build", "{projectRoot}/tsconfig*.tsbuildinfo"]
     },
     "typecheck": {
       "dependsOn": ["^build"]
     },
     "test": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["build"],
```
oleg-semyonov, line 19: Please check that tests for `theme-kit-nextjs` work with build directory cleaned. Since it depends on `theme-kit-core`, it requires both libraries to be built prior to testing

oleg-semyonov (commented):
Looks good, got only one concern on the NX config

nx.json
e1himself: Yes, it does work. 

This is because `test` is set to depend on same-package `build` task, while the `build` task itself is set to depend on `build` output of its dependencies (`^build`).

These are the tasks that are executed when `npm run test` is invoked in the repo root:

    ✔  @prezly/theme-kit-core:prebuild (505ms)
    ✔  @prezly/theme-kit-nextjs:prebuild (524ms)
    ✔  @prezly/theme-kit-intl:prebuild (2s)
    ✔  @prezly/theme-kit-core:build (2s)
    ✔  @prezly/theme-kit-core:test (4s)
    ✔  @prezly/theme-kit-nextjs:build (5s)
    ✔  @prezly/theme-kit-intl:build (9s)
    ✔  @prezly/theme-kit-nextjs:test (4s)
    ✔  @prezly/theme-kit-intl:test (10s)

e1himself: This is also proven by the Github Actions test workflow:

https://github.com/prezly/theme-kit-js/blob/8e779cb4d33df0ee99bdd041079bfb0de6b2f7fe/.github/workflows/test.yml#L49-L50

It never calls `build` prior to invoking `npm test`

oleg-semyonov: Great, thanks for double-checking!

oleg-semyonov (approved):

PR #404, opened 2023-10-19
[DEV-11819] Add buttons components to Storybook

fgyimah:


packages/ui/.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
         project: ['./tsconfig.eslint.json'],
         tsconfigRootDir: __dirname,
     },
-    rules: {},
+    rules: { 'func-style': 'off' },
```
oleg-semyonov, line 7: Let's not disable this rule for the whole package. You can disable the rule for specific line, and that is okay in your use-case (forwardRef)

packages/ui/src/components/Button/Icon.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export function Icon({ icon: IconComponent, placement, iconOnly }: Props) {
+    const isLeft = placement === 'left';
+
+    if (IconComponent) {
+        return (
+            <IconComponent
+                width={16}
+                height={16}
+                className={twMerge('w-[1em] h-[1em]', !iconOnly && (isLeft ? 'mr-2' : 'ml-2'))}
```
oleg-semyonov, line 19: It's better to use flex gap on the button container itself, rather than margins on the icons.

packages/ui/src/components/Button/Icon.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface Props {
+    icon?: BaseProps['icon'];
+    placement: 'left' | 'right';
+    iconOnly?: boolean;
+}
+
+export function Icon({ icon: IconComponent, placement, iconOnly }: Props) {
+    const isLeft = placement === 'left';
+
+    if (IconComponent) {
```
oleg-semyonov, line 14: Better to return null early

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
+import type { ButtonHTMLAttributes, PropsWithChildren } from 'react';
+import { forwardRef } from 'react';
+import { twMerge } from 'tailwind-merge';
+
+import { Icon } from './Icon';
+import type { BaseProps } from './types';
+
+export interface ButtonProps extends BaseProps, ButtonHTMLAttributes<HTMLButtonElement> {
```
oleg-semyonov, line 8: Shouldn't we use namespace to exports props like we do in other projects? 🤔 

This is a preference thing, so would like to hear @e1himself's opinion on that

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+                `,
+                Boolean(icon) && !children && `ml-0 mr-0`,
+            )}
+            onClick={onClick}
+            disabled={isDisabled || isLoading}
+            {...buttonProps}
+        >
+            {iconPlacement === 'left' && <Icon icon={icon} iconOnly={!children} placement="left" />}
+            {/* If there are no children, we insert a zero-width space to preserve the line-height */}
+            <span className={contentClassName}>{children ?? <>&#8203;</>}</span>
```
oleg-semyonov, line 70: Switching to flex-gap as suggested in my previous comment would mess up this line-height hack, but you can add the gap property conditionally - only if button is not `iconOnly`

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
+import type { ButtonHTMLAttributes, PropsWithChildren } from 'react';
+import { forwardRef } from 'react';
+import { twMerge } from 'tailwind-merge';
+
+import { Icon } from './Icon';
+import type { BaseProps } from './types';
+
+export interface ButtonProps extends BaseProps, ButtonHTMLAttributes<HTMLButtonElement> {
+    isLoading?: boolean;
+    isDisabled?: boolean;
```
oleg-semyonov, line 10: There's already a `disabled` property on HTMLButton, let's re-use that.
Same for `onClick`.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+    onClick?: () => void;
+    contentClassName?: string;
+}
+
+export const Button = forwardRef<
+    HTMLButtonElement,
+    Omit<PropsWithChildren<ButtonProps>, 'onResize' | 'onResizeCapture'>
```
oleg-semyonov, line 17: Shouldn't these properties be omitted from `ButtonHTMLAttributes` instead?

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+                variation === 'secondary' &&
+                    `
+                    bg-white border-gray-200 text-gray-800
+                hover:border-gray-300 hover:bg-gray-50
+                focus:bg-gray-50 focus:border-2 focus:border-[color:var(--prezly-accent-color-lighter)]
+                focus-within:bg-gray-50 focus-within:border-2 focus-within:border-[color:var(--prezly-accent-color-lighter)]
+                active:bg-gray-100 active:border-gray-400
+                disabled:bg-white disabled:border-transparent disabled:text-gray-400
+                `,
+                Boolean(icon) && !children && `ml-0 mr-0`,
```
oleg-semyonov, line 62: I don't get what is this for. Button should not have any margin by default anyway 🤔

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { PlusIcon } from '@heroicons/react/24/solid';
```
oleg-semyonov, line 2: So the heroicons are only used in the stories? Then it should be a dev dependency or optional dependency

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+                </div>
+            ))}
+        </>
+    );
+};
+
+AllVariations.parameters = {
+    design: {
+        type: 'figma',
+        url: 'https://www.figma.com/file/46dEAasj1iEtrVQOwmMswB/00--%3E-Themes-Design-System?type=design&node-id=601-3728&mode=dev',
```
oleg-semyonov, line 78: It seems like the Figma plugin was not installed, there's no `Design` tab in the story.

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { Button, type ButtonProps } from '.';
+
+const Icon = PlusIcon;
+
+export default {
+    title: 'Buttons/Button',
+    component: Button,
+} as Meta<typeof Button>;
+
+export const AllVariations: StoryFn<typeof Button> = (args) => {
```
oleg-semyonov, line 16: Do we really need this story? You can still play around with props on any story. Also, this story doesn't even fit on a screen, so it's hard to overview or snapshot test anyway.

I'd rather make a couple extra stories for disabled/loading state. You can also use `autodocs` to generate an overview page with all stories.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+            ...buttonProps
+        },
+        ref,
+    ) => (
+        <button
+            ref={ref}
+            // eslint-disable-next-line react/button-has-type
+            type={type}
+            className={twMerge(
+                'rounded-[4px] flex items-center justify-center border-[1px] border-transparent bg-transparent',
```
oleg-semyonov, line 41: You're using the default Tailwind values, so no need for arbitrary syntax

```suggestion
                'rounded flex items-center justify-center border border-transparent bg-transparent',
```

oleg-semyonov (changes requested):

packages/ui/src/components/Button/Button.tsx
e1himself: I do prefer using namespaces for these: fewer symbols to export.

```ts
import { Button } from '__';

interface Props extends Button.Props {
  // ...
}
```

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+                focus:bg-[color:var(--prezly-accent-color-dark)] focus:border-2 focus:border-[color:var(--prezly-accent-color-lighter)]
+                focus-within:bg-[color:var(--prezly-accent-color-dark)] focus-within:border-2 focus-within:border-[color:var(--prezly-accent-color-lighter)]
+                active:bg-[color:var(--prezly-accent-color-darker)] active:border-[color:var(--prezly-accent-color-darker)]
+                disabled:bg-[color:var(--prezly-accent-color-lighter)] disabled:border-transparent
+                `,
+                variation === 'secondary' &&
+                    `
+                    bg-white border-gray-200 text-gray-800
+                hover:border-gray-300 hover:bg-gray-50
+                focus:bg-gray-50 focus:border-2 focus:border-[color:var(--prezly-accent-color-lighter)]
```
e1himself, line 57: Border width changes here, causing layout shifts:

[Screencast from 2023-10-20 12-40-58.webm](https://github.com/prezly/theme-kit-js/assets/370680/67c92c94-7353-44b0-8bfd-38cc140a157e)

oleg-semyonov: If you want a secondary border on focus, use `ring` classes for that

packages/ui/src/components/Button/Button.tsx
fgyimah: I'm unable to export namespace with same name as a `const` variable

e1himself: Yeah, that's indeed unfortunate.

The best workaround we've found so far is to use `forwardRef` property instead of React's stock `ref`:

https://github.com/prezly/prezly/blob/39fc1fc99fffd5dbb51ec74a3a15adaa0b46c0a2/apps/backend/resources/javascripts/ui/Button/Button.tsx#L21

fgyimah: thanks @e1himself , will update to use this approach

fgyimah: Updated https://github.com/prezly/theme-kit-js/pull/404/commits/25d3d48a9248d6c224e985e7870e5f1681153016

packages/ui/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return (
+        <button
+            ref={forwardRef}
+            type={type}
+            className={twMerge(
+                'rounded flex items-center justify-center border border-transparent bg-transparent',
+                size === 'small' ? `p-3 label-medium` : `p-4 label-large`,
+                variation === 'primary' &&
+                    `
+                bg-[color:var(--prezly-accent-color)] border-[color:var(--prezly-accent-color)] text-[color:var(--prezly-accent-color-button-text)]
```
e1himself, line 30: > `bg-[color:var(--prezly-accent-color)]`

I wonder if we can define these as tailwind classes.  E.g. `bg-accent-color`?

This is possible? @oleg-semyonov wdyt?

fgyimah: Updated to use css variables in the tailwind config

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+export const SecondaryWithIconLeft = ButtonTemplateWithChild.bind({});
+SecondaryWithIconLeft.args = {
+    ...Secondary.args,
+    icon: Icon,
+    iconPlacement: 'left',
+};
+
+export const SquareButton = ButtonTemplateWithoutChild.bind({});
```
oleg-semyonov, line 49: Square button is not exactly square 😉 

image

oleg-semyonov: Also, I think the line height for the button is weird, I was expecting round values for text line-height, at least for buttons.
image

oleg-semyonov (changes requested):
Looking good, except one minor thing with line-height 🤔

packages/ui/src/components/Button/Button.stories.tsx
fgyimah: Updated https://github.com/prezly/theme-kit-js/pull/404/commits/f4b350576eda805db5368a8be746d725d349af4c

packages/ui/src/typography.styles.css
```diff
@@ -51,7 +51,7 @@ body {
 }
 
 .label-large {
-    @apply text-sm font-normal leading-[160%] text-gray-800 sm:text-base sm:font-medium sm:leading-[124%];
+    @apply text-sm font-normal leading-[160%] text-gray-800 sm:text-base sm:font-medium sm:leading-[20px];
```
oleg-semyonov, line 54: I guess it's better to still use percent values for better scalability, here it would be 125%

packages/ui/src/components/Button/Button.tsx
```diff
@@ -44,6 +44,7 @@ export function Button({
                 disabled:bg-white disabled:border-transparent disabled:text-gray-400
                 `,
                 Boolean(Icon) && Boolean(children) && `gap-2`,
+                Boolean(Icon) && !children && `px-4 py-3`,
```
oleg-semyonov, line 47: The button is still not squared, but now in another way :D

image

oleg-semyonov (changes requested):
Sorry, not quite there yet 😉

packages/ui/src/components/Button/Button.tsx
fgyimah: Fixed it. https://github.com/prezly/theme-kit-js/pull/404/commits/4f07d07b37ab7c8da93e1458eb4a0871d3147dd7

packages/ui/src/typography.styles.css
fgyimah: https://github.com/prezly/theme-kit-js/pull/404/commits/c3ea8f70811b98b5dc627d4a305b2537ababbcbc

packages/ui/src/components/Button/Button.tsx
```diff
@@ -44,15 +44,13 @@ export function Button({
...
                 Boolean(Icon) && Boolean(children) && `gap-2`,
-                Boolean(Icon) && !children && `px-4 py-3`,
                 className,
             )}
             disabled={disabled || isLoading}
             {...buttonProps}
         >
             {iconPlacement === 'left' && <Icon icon={icon} />}
-            {/* If there are no children, we insert a zero-width space to preserve the line-height */}
-            <span className={contentClassName}>{children ?? <>&#8203;</>}</span>
```
oleg-semyonov: Removing this fix made the square button smaller than regular button (50px vs 54px).
Please test more carefully when fixing UI bugs 🙏🏻

oleg-semyonov (changes requested):
Sorry, still can't approve yet 😕

packages/ui/src/components/Button/Button.tsx
fgyimah: Fixed https://github.com/prezly/theme-kit-js/pull/404/commits/4429bf7ad83307b0cdc2850c7b4d3bc846b2f799

PR #414, opened 2023-10-24
[DEV-11849] & [DEV-11850] Add Hero and StoryCard component to storybook

fgyimah:


packages/ui/src/components/StoryCard/StoryCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+                size === 'large' ? `flex-col gap-0` : `md:flex-row items-start gap-4`,
+                className,
+            )}
+        >
+            <Link href={`/${slug}`} locale={false} legacyBehavior>
+                <div
+                    className={twMerge(
+                        size === 'large'
+                            ? 'w-full aspect-[20/14] md:aspect-[27/17]'
+                            : 'w-1/3 aspect-[10/7] hidden md:block',
```
e1himself, line 48: Isn't `10/7` the same as `20/14`?

packages/ui/src/components/StoryPublicationDate/StoryPublicationDate.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function StoryPublicationDate({ story, locale }: Props) {
+    const date = getStoryPublicationDate(story);
+
+    if (!date) {
+        return null;
+    }
+
+    return (
+        <>{date.toLocaleDateString(locale, { year: 'numeric', month: 'long', day: 'numeric' })}</>
+    );
```
e1himself, line 18: I believe date formats should be customizable. 

We have this setting available in the app:

image

packages/ui/src/components/CategoryLink/CategoryLink.tsx
```diff
@@ -0,0 +1,23 @@
+import type { Category, Culture } from '@prezly/sdk';
+import { getCategoryUrl, getLocalizedCategoryData, LocaleObject } from '@prezly/theme-kit-core';
+import Link from 'next/link';
+import { twMerge } from 'tailwind-merge';
+
+interface Props {
+    category: Category;
```
e1himself, line 7: It is recommended to `Pick<...>` only the properties the component needs to make it possible using the library with a wider range of `@prezly/sdk` package versions.

Read more about it here: https://github.com/prezly/theme-kit-js#prezly-sdk-dependency

packages/ui/src/components/StoryCard/StoryCard.tsx
fgyimah: Is the same, but the large has a different aspect ration from tablet to desktop viewport

e1himself: I'm just wondering why the same aspect ratio is used with two different notations. `20/14` can be replaced with `10/7` with no side effects. Right?

fgyimah: Yeah right

fgyimah: Updated https://github.com/prezly/theme-kit-js/pull/414/commits/7f5e4071f2ebd39b7cae32883fec555728b1ac8b

packages/ui/src/components/StoryPublicationDate/StoryPublicationDate.tsx
fgyimah: Updated https://github.com/prezly/theme-kit-js/pull/414/commits/7f5e4071f2ebd39b7cae32883fec555728b1ac8b

packages/ui/src/components/CategoryLink/CategoryLink.tsx
fgyimah: Updated https://github.com/prezly/theme-kit-js/pull/414/commits/7f5e4071f2ebd39b7cae32883fec555728b1ac8b

packages/ui/src/types.ts
```diff
@@ -0,0 +1,3 @@
+import type { Story } from '@prezly/sdk';
+
+export type StoryWithImage = Story & Pick<Story.ExtraFields, 'thumbnail_image'>;
```
oleg-semyonov, line 3: Hey @e1himself, I wonder if we should include the `thumbnail_image` into base `Story` object at some point. We're using this field in every theme, and this type we've been using feels a bit like a hack 🤔

oleg-semyonov (commented):
Looks good! 👍🏻

packages/ui/src/types.ts
e1himself: >  I wonder if we should include the thumbnail_image into base Story object at some point. 

Hey Oleg! 

I'm not sure yet. The thumb rule for picking defaults for Stories API response fields is whether it is needed on the listing page (in the app), and `thumbnail_image` is not needed there.

But I see your point. I'll need to think about it.

PR #423, opened 2023-10-31
[DEV-11883] Add Navigation component to Storybook

fgyimah:


packages/ui/src/hooks/useDevice.ts
```diff
@@ -0,0 +1,18 @@
...
+const BREAKPOINT_TABLET = 414;
+const BREAKPOINT_DESKTOP = 768;
+
+export function useDevice() {
+    const isMobile = useMediaQuery(`(max-width: ${BREAKPOINT_TABLET}px)`, {
+        initializeWithValue: true,
+    });
+    const isTablet = useMediaQuery(`(max-width: ${BREAKPOINT_DESKTOP}px)`, {
+        initializeWithValue: true,
+    });
```
oleg-semyonov, line 12: IMO It's better to sync the names and breakpoints with Tailwind, otherwise it would get pretty confusing 🤔

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,207 @@
...
+                        </div>
+                    </div>
+                </div>
+                <div className="flex items-center gap-4 md:hidden">
+                    {Boolean(isSearchEnabled && onSearch) && (
+                        <Button
+                            className="p-0"
+                            variation="navigation"
+                            icon={MagnifyingGlassIcon}
+                            iconClassName="w-6 h-6"
```
oleg-semyonov, line 192: This feels incorrect. 
You're applying the same `className` and `iconClassName` to all navigation buttons. These styles should be part of the `navigation` variant and attached in the `Button` component, otherwise we're missing the whole point of ready-made UI components.

`iconClassName` should not be exposed from the `Button` component at all IMO.

packages/ui/src/components/Navigation/Navigation.stories.tsx
```diff
@@ -0,0 +1,880 @@
...
+
+export default {
+    title: 'Components/Navigation',
+    component: Navigation,
+    parameters: { layout: 'fullscreen' },
+} as Meta<typeof Navigation>;
+
+const NavigationTemplate: StoryFn<typeof Navigation> = (args) => <Navigation {...args} />;
+
+export const Default = NavigationTemplate.bind({});
```
oleg-semyonov, line 16: Please add more stories with different variations.
There are two layouts, so that's two stories already.
Then it's good to show how would navigation look without categories, without an external link, without languages, without site logo, etc.

packages/ui/src/components/Navigation/Navigation.stories.tsx
```diff
@@ -0,0 +1,880 @@
...
+Default.args = {
+    categories: [
```
oleg-semyonov, line 18: This is better extracted into a mocks file.

packages/ui/src/components/Navigation/Navigation.stories.tsx
```diff
@@ -0,0 +1,880 @@
...
+                        language_code: 'en',
+                    },
+                    name: 'Family',
+                    description: '',
+                    slug: 'family',
+                },
+            },
+        },
+    ] as Category[],
+    languages: [
```
oleg-semyonov, line 286: Same here, these objects are too large and make stories unreadable. Please extract into a separate mock file.

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,207 @@
...
+export interface Props {
+    className?: string;
+    categories: Category[];
+    languages: NewsroomLanguageSettings[];
+    layout?: NavigationLayout;
+    logo?: UploadedImage | null;
+    siteName: string;
+    publicGalleriesCount: number;
+    externalSiteLink?: string;
+    isSearchEnabled?: boolean;
```
oleg-semyonov, line 36: I believe that all of these properties can be pulled from `Newsroom` API object. That would help reduce the amount of props required.

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,207 @@
...
+    onSearch?: () => void;
+    locale: Culture['code'];
+    hasStandaloneAboutPage?: boolean;
+    hasStandaloneContactsPage?: boolean;
+    hasError?: boolean;
+    currentStory?: ExtendedStory;
+    currentCategory?: Pick<Category, 'i18n' | 'display_name'>;
```
oleg-semyonov, line 43: These props should not be optional, since they're required for Language Picker to work properly.
```suggestion
    currentStory: ExtendedStory | undefined;
    currentCategory: Pick<Category, 'i18n' | 'display_name'> | undefined;
```

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,207 @@
+import {
```
oleg-semyonov, line 1: Please don't forget to add `use client` directive for components that are using state, hooks or event handlers

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,207 @@
...
+                                locale={locale}
+                            >
+                                {/* TODO: Use translations */}
+                                Contacts
+                            </Link>
+                        )}
+                    </div>
+
+                    <div className="flex flex-col md:flex-row md:items-center gap-12 md:gap-4">
+                        {Boolean(isSearchEnabled && onSearch) && (
```
oleg-semyonov, line 151: I wonder if we can combine these two props into one?
If you don't pass `onSearch` handler, the search would not be enabled anyway, so that makes `isSearchEnabled` prop kind of redundant 🤔

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,207 @@
...
+                                    className="label-large text-gray-600 hover:text-gray-800 flex items-center shrink-0"
+                                    href={externalSiteLink}
+                                    target="_blank"
+                                    rel="noopener noreferrer"
+                                    data-title={
+                                        <>
+                                            {extractDomainFromUrl(externalSiteLink)}
+                                            <ArrowUpRightIcon className="ml-1 w-2 h-2" />
+                                        </>
+                                    }
```
oleg-semyonov, line 177: What is this needed for? The text content of the link is already picked up by assistive technologies. Also, you can't put JSX into data-attributes anyway 🤔

packages/ui/src/components/Navigation/components/LanguagesDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+            className="border-0 w-max p-0 bg-transparent text-gray-600 hover:text-gray-800"
+            contentClassName="border mt-2 p-0"
+            label={<GlobeAltIcon className="w-[20px] h-[20px]" />}
+        >
+            <Dropdown.Group>
+                {displayedLanguages.map((language, index) => {
+                    const translationLink = hasError
+                        ? '/'
+                        : getTranslationUrl(
+                              LocaleObject.fromAnyCode(language.code),
```
oleg-semyonov, line 51: Shouldn't new components use the new `Locale` object created by @e1himself? 🤔

packages/ui/src/components/Navigation/util.ts
```diff
@@ -0,0 +1,89 @@
+import type { Category, ExtendedStory } from '@prezly/sdk';
+import {
+    getCategoryHasTranslation,
+    getCategoryUrl,
+    isStoryPublished,
+    type LocaleObject,
+    Visibility,
+} from '@prezly/theme-kit-core';
+
+export function extractDomainFromUrl(url: string): string {
```
oleg-semyonov, line 10: All of this code can be replaced with just one line :)
```ts
return new URL(url).hostname;
```

packages/ui/src/components/Navigation/components/CategoriesDropdown.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { Button } from '@/components/Button';
+import { Dropdown } from '@/components/Dropdown';
+import { useDevice } from '@/hooks';
+
+export interface Props {
+    categories: Category[];
+    locale: Culture['code'];
+}
+
+export function CategoriesDropdown({ categories, locale }: Props) {
```
oleg-semyonov, line 17: IMO, the Categories on desktop look real bad :(
Screenshot 2023-11-02 at 11 54 50

Could you advise with Al/Hugo on how this can be improved?

packages/ui/src/components/Navigation/components/LanguagesDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export interface Props {
+    languages: NewsroomLanguageSettings[];
+    locale: Culture['code'];
+    currentStory?: ExtendedStory;
+    currentCategory?: Pick<Category, 'i18n' | 'display_name'>;
+    hasError?: boolean;
+}
+
+export function LanguagesDropdown({
```
oleg-semyonov, line 19: Languages dropdown is almost inaccessible on mobile screens:

image

oleg-semyonov (changes requested):
Sorry, can't approve in this state. Please see my comments.

packages/ui/src/components/Navigation/components/CategoriesDropdown.tsx
fgyimah: Updated

packages/ui/package.json
```diff
@@ -65,9 +65,11 @@
   },
   "dependencies": {
     "@heroicons/react": "^2.0.18",
-    "@prezly/theme-kit-core": "^6.6.0",
+    "@prezly/theme-kit-core": "0.0.0-next.3+5d656f7",
+    "@prezly/theme-kit-intl": "0.0.0-next-22e966191ad837b196b46db94834a1eade540e17",
```
oleg-semyonov, line 69: Okay, I got a bit confused about different branches. I can see now that the new Local object is part of `next` branch, which is not yet released.
Since you're merging into the `main` branch, you shouldn't depend on future unreleased versions.

Sorry for confusion, my comment about the `Locale` object should be disregarded. Let's use current public APIs until the `next` branch is merged into `main`

packages/ui/src/components/Navigation/Navigation.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+            version: 2,
+            uuid: '13b5794c-619e-4c6f-961c-0b8e15d771c9',
+            filename: 'goodnewsroom-new.png',
+            mime_type: 'image/png',
+            size: 21659,
+            original_width: 916,
+            original_height: 331,
+            effects: [],
+        },
+    },
```
oleg-semyonov, line 65: You can re-use args from other stories to reduce boilerplate:
```ts
Centered.args = {
    ...Default.args,
    layout: 'centered',
};
```

oleg-semyonov: Also the newsroom could be extracted into mocks as well

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,220 @@
...
+                                    {/* TODO: Use translations */}
+                                    Contacts
+                                </Link>
+                            )}
+                        </div>
+                    )}
+
+                    <div className="flex flex-col md:flex-row md:items-center gap-12 md:gap-4">
+                        {Boolean(onSearch) && (
+                            <a className="hidden md:flex" href="#" onClick={handleSearch}>
```
oleg-semyonov, line 171: Why is this a link and not a button? You're using the button in the mobile markup below 🤔

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,220 @@
...
+                                    {/* TODO: Use translations */}
+                                    About
+                                </Link>
+                            )}
+                            {hasStandaloneContactsPage && (
+                                <Link
+                                    className={twMerge(
+                                        'label-large text-gray-600 hover:text-gray-800 shrink-0',
+                                        !isSm && `text-lg font-bold`,
+                                    )}
```
oleg-semyonov, line 158: This class names construction is re-used multiple times, so probably better to extract the link into a component, or compute the class name once somewhere up in the code.

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,220 @@
...
+                                    href="/contacts"
+                                    locale={locale}
```
oleg-semyonov, line 160: All other links have `locale={locale ?? false}`, but not this one. I guess that was not intended?

packages/ui/src/hooks/useDevice.ts
```diff
@@ -0,0 +1,27 @@
+import { useMediaQuery } from '@react-hookz/web';
+
+const BREAKPOINT_SM = 640;
+const BREAKPOINT_MD = 768;
+const BREAKPOINT_LG = 1024;
+const BREAKPOINT_XL = 1280;
+const BREAKPOINT_2XL = 1536;
```
oleg-semyonov, line 7: I saw you were importing colors from Tailwind in a Storybook doc, can't you do the same for breakpoints? That would make the maintenance much easier, as you won't have to keep these values in sync manually.

oleg-semyonov (changes requested):
Looks much better now!

Still, got a couple more remarks.

packages/ui/src/hooks/useDevice.ts
fgyimah: This required importing the `tailwind.config.js` and resolving with `tailwindcss/resolveConfig`, it brought up issues around `module.exports` used in the tailwind.config file and changing to export default caused tailwind to use default config (not able to use custom config)

oleg-semyonov: Okay, thanks for checking this out 👍🏻 
Let's just remember to keep those values in sync. Maybe worth adding a comment in `tailwind.config.js` that values in `useDevices` need to be kept in sync with the config.

oleg-semyonov (approved):
Great work! 👏🏻

packages/ui/src/hooks/useDevice.ts
fgyimah: https://github.com/prezly/theme-kit-js/pull/423/commits/636e7d419d73a41752cf4e1e134276da2532a013

PR #424, opened 2023-11-01
[DEV-11899] Add Links and Badges to Storybook

fgyimah:


packages/ui/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <div className="bg-gray-100 p-10">
+                <Story />
+            </div>
+        ),
+    ],
+} as Meta<typeof Badge>;
+
+const BadgeTemplateWithChild: StoryFn<typeof Badge> = (args) => <Badge {...args}>Badge</Badge>;
+
+export const Primary = BadgeTemplateWithChild.bind({});
```
oleg-semyonov, line 22: Please also add stories for `small` variation

packages/ui/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import { ChevronDownIcon } from '@heroicons/react/24/solid';
+import type { Meta, StoryFn } from '@storybook/react';
+
+import { Badge } from './Badge';
+
+const Icon = ChevronDownIcon;
```
oleg-semyonov, line 6: Can we use another icon here? It makes the UI kinda confusing (looks like a dropdown)

packages/ui/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export default {
+    title: 'Buttons/Badge',
```
oleg-semyonov, line 9: I wonder, if this component is under `Button` section, acts like a button, and has all of the same layout and styles, why can't this just be a prop of the `Button` component? (e.g. `rounded`).

If it walks like a duck and quacks like a duck.. 😉

packages/ui/src/components/Button/ButtonLink.tsx
```diff
@@ -40,7 +40,7 @@ export function ButtonLink({
   focus:bg-accent-dark focus:border-transparent focus:ring-2 focus:ring-accent-lighter
   focus-within:bg-accent-dark focus-within:border-transparent focus-within:ring-2 focus-within:ring-accent-lighter
   active:bg-accent-dark active:border-accent-dark
-  disabled:bg-accent-lighter disabled:border-transparent
+  ${disabled && `bg-accent-lighter border-transparent pointer-events-none opacity-50`}
```
oleg-semyonov, line 43: No need to use template string quotes when template syntax is not used. Same for line 52.
```suggestion
  ${disabled && 'bg-accent-lighter border-transparent pointer-events-none opacity-50'}
```

packages/ui/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,77 @@
...
+    disabled,
+    children,
+    contentClassName,
+    localeCode,
+    href,
+    ...linkProps
+}: Link.Props) {
+    const Element = forceRefresh ? 'a' : NextLink;
+
+    const isIconOnly = Boolean(Icon) && !children;
```
oleg-semyonov, line 25: I don't think a Link can be icon-only. That is true for buttons, but links are usually intended to be placed inside text, so an icon-only Link would be really weird to see.

oleg-semyonov (changes requested):

packages/ui/src/components/Badge/Badge.stories.tsx
fgyimah: Will make this a prop of the button

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -64,3 +64,33 @@ LoadingButton.args = {
     icon: Icon,
     isLoading: true,
 };
+
+export const Badge = ButtonTemplateWithChild.bind({});
```
oleg-semyonov, line 68: Let's also rename the stories to `Rounded` so that it's clear how its connected to the prop.
I believe the naming in the Figma designs should also be updated.

oleg-semyonov: By the way, shouldn't these stories be connected to Figma designs? 😉

packages/ui/src/components/Button/Button.tsx
```diff
@@ -25,23 +26,35 @@ export function Button({
...
             type={type}
             className={twMerge(
-                'rounded flex items-center justify-center border border-transparent bg-transparent',
-                size === 'small' ? `p-3 label-medium` : `p-4 label-large`,
+                `${
+                    rounded ? 'rounded-full' : 'rounded'
+                } flex items-center justify-center border border-transparent bg-transparent`,
+                size === 'small'
+                    ? `${rounded ? 'py-2 px-3' : 'p-3'} label-medium`
+                    : `${rounded ? 'py-3 px-4' : 'p-4'} label-large`,
```
oleg-semyonov, line 34: I know this is by Figma design, but it feels weird that spacing is different for rounded buttons.
I'd consult about that with Hugo and Al. Ideally we'd want the spacing to be uniform. If we need extra variations, it's better to add more sizing variants.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -25,23 +26,35 @@ export function Button({
...
                 variation === 'primary' &&
                     `
                 bg-accent border-accent text-accent-button-text
                 hover:bg-accent-dark hover:border-accent-dark 
-                focus:bg-accent-dark focus:border-transparent focus:ring-2 focus:ring-accent-lighter
-                focus-within:bg-accent-dark focus-within:border-transparent focus-within:ring-2 focus-within:ring-accent-lighter
+                focus:bg-accent-dark focus:border-transparent ${
+                    rounded
+                        ? 'focus:ring-4 focus-within:ring-4'
+                        : 'focus:ring-2 focus-within:ring-2'
```
oleg-semyonov, line 42: I don't get why the focus styling should be different for the rounded button? 
Currently the focus ring for regular button feels too thing, I'd use the `focus:ring-4` for all variants.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -25,23 +26,35 @@ export function Button({
...
                     `
-                    bg-white border-gray-200 text-gray-800
+                    bg-white ${rounded ? 'border-transparent' : 'border-gray-200'} text-gray-800
                 hover:border-gray-300 hover:bg-gray-50
-                focus:bg-gray-50 focus:border-transparent focus:ring-2 focus:ring-accent-lighter
-                focus-within:bg-gray-50 focus-within:border-transparent focus-within:ring-2 focus-within:ring-accent-lighter
+                focus:bg-gray-50 focus:border-transparent ${
+                    rounded
+                        ? 'focus:ring-4 focus-within:ring-4'
+                        : 'focus:ring-2 focus-within:ring-2'
```
oleg-semyonov, line 55: These styles are identical to `primary` variant, so it's probably best to move them to common styles.

packages/ui/src/components/Button/Button.tsx
```diff
@@ -25,23 +26,35 @@ export function Button({
...
+                }  focus:ring-accent-lighter
+                focus-within:bg-accent-dark focus-within:border-transparent focus-within:ring-accent-lighter
                 active:bg-accent-dark active:border-accent-dark
-                disabled:bg-accent-lighter disabled:border-transparent
+                disabled:bg-accent-lighter disabled:border-transparent disabled:opacity-50
                 `,
                 variation === 'secondary' &&
                     `
-                    bg-white border-gray-200 text-gray-800
+                    bg-white ${rounded ? 'border-transparent' : 'border-gray-200'} text-gray-800
```
oleg-semyonov, line 50: Again, it feels weird that colors of the button change based on the `rounded` property. This is another thing to discuss with the design team.

packages/ui/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+  `,
+                Boolean(Icon) && Boolean(children) && `gap-2`,
+                className,
+            )}
+            {...linkProps}
+        >
+            {iconPlacement === 'left' && <Icon icon={icon} />}
+            {/* If there are no children, we insert a zero-width space to preserve the line-height */}
+            <span className={contentClassName}>{children ?? <>&#8203;</>}</span>
```
oleg-semyonov, line 54: Based on our previous conversation, `children` should be a required prop, and this hack becomes redundant.

oleg-semyonov (changes requested):
This looks cleaner, but I think you need to consult with Al and Hugo to align the designs and the implementation. Currently it's still confusing (`rounded` prop somehow also controls spacing and border color)

packages/ui/src/components/Button/Button.stories.tsx
fgyimah: Updated

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -70,3 +70,34 @@ LoadingButton.args = {
     icon: Icon,
     isLoading: true,
 };
+
+export const Rounded = ButtonTemplateWithChild.bind({});
+Rounded.parameters = {
+    design: {
+        type: 'figma',
+        url: 'https://www.figma.com/file/46dEAasj1iEtrVQOwmMswB/00--%3E-Themes-Design-System?type=design&node-id=1128-39276&mode=dev',
```
oleg-semyonov, line 78: `Design` tab is not present in Storybook, maybe some config is set up incorrectly? Could you check please?

packages/ui/src/components/Button/Button.tsx
```diff
@@ -16,32 +16,33 @@ export function Button({
...
     return (
         <button
             ref={forwardRef}
             type={type}
             className={twMerge(
-                'rounded flex items-center justify-center border border-transparent bg-transparent',
-                size === 'small' ? `p-3 label-medium` : `p-4 label-large`,
+                `${
+                    rounded ? 'rounded-full' : 'rounded'
+                } flex items-center justify-center border border-transparent bg-transparent focus:ring-4 focus-within:ring-4 focus:ring-accent-lighter focus-within:ring-accent-lighter`,
```
oleg-semyonov, line 29: It would be cleaner to place this tertiary in a separate string like you do with other conditional properties, that way it would be placed on a single line
```suggestion
                rounded ? 'rounded-full' : 'rounded',
                'flex items-center justify-center border border-transparent bg-transparent focus:ring-4 focus-within:ring-4 focus:ring-accent-lighter focus-within:ring-accent-lighter',
```

packages/ui/src/components/Button/Button.tsx
```diff
@@ -59,12 +60,14 @@ export function Button({
             {iconPlacement === 'left' && (
                 <Icon
                     icon={icon}
-                    className={twMerge(isIconOnly && (size === 'small' ? `w-4 h-4` : `w-6 h-6`))}
+                    className={twMerge(
+                        !children && variation !== 'navigation' ? 'w-4 h-6' : 'w-6 h-6',
```
oleg-semyonov, line 64: Shouldn't it be `w-4 h-4` ?
```suggestion
                        !children && variation !== 'navigation' ? 'w-4 h-4' : 'w-6 h-6',
```

packages/ui/src/components/Button/Button.tsx
```diff
@@ -59,12 +60,14 @@ export function Button({
...
+                    )}
                 />
             )}
             {/* If there are no children, we insert a zero-width space to preserve the line-height */}
             <span className={contentClassName}>{children ?? <>&#8203;</>}</span>
-            {iconPlacement === 'right' && <Icon icon={icon} />}
+            {iconPlacement === 'right' && <Icon icon={icon} className="w-4 h-6" />}
```
oleg-semyonov, line 70: I guess this should receive the same class name as the left icon on line 63

packages/ui/src/components/Button/Button.tsx
fgyimah: It's `w-4 h-6` to make it a square button because of the padding on the container

packages/ui/src/components/Button/Button.stories.tsx
fgyimah: Fixed this

packages/ui/src/components/Button/Button.tsx
oleg-semyonov: Doesn't the zero-width space serve this purpose? It was added there exactly for this reason 🤔

fgyimah: Yeah but button has padding `12px 16px` so using rounded values for height and width gives a rectangular button instead of square button

oleg-semyonov: Oh, right. Thanks for clearing this one up! 👍🏻

fgyimah: Added comment

oleg-semyonov (approved):
👏🏻

PR #425, opened 2023-11-03
[DEV-11852] Add Subscription CTA to storybook

fgyimah:


packages/ui/src/components/SubscriptionCta/SubscriptionCta.tsx
```diff
@@ -0,0 +1,48 @@
...
+                onSubmit={handleSubmit}
+            >
+                <Input
+                    className="w-full"
+                    inputClassName="w-full"
+                    //  TODO: add translations
+                    placeholder="Enter your email"
+                    value={value}
+                    onChange={(event) => onChange(event.target.value)}
+                />
```
oleg-semyonov, line 41: There should be an error state (e.g. invalid email) supported in this component.

packages/ui/src/components/SubscriptionCta/SubscriptionCta.tsx
```diff
@@ -0,0 +1,48 @@
...
+        >
+            {/* TODO: add translations */}
+            <h3 className="title-large w-full md:w-1/2 text-white">Get updates in your inbox</h3>
+            <form
+                className="flex flex-col sm:flex-row items-center w-full md:w-1/2 gap-4"
+                onSubmit={handleSubmit}
+            >
+                <Input
+                    className="w-full"
+                    inputClassName="w-full"
```
oleg-semyonov, line 36: Shouldn't input always be the full width of its container? IMO this should be encapsulated in the `Input` component, then you'll only control the overall component size.

oleg-semyonov (changes requested):

oleg-semyonov (approved):
Please also add a story with an error, then good to merge 👍🏻

PR #436, opened 2023-11-06
[DEV-11853] Add Footer to Storybook

fgyimah:


packages/ui/src/components/Boilerplate/Boilerplate.tsx
```diff
@@ -0,0 +1,80 @@
...
+                            >
+                                {email}
+                            </Link>
+                        </div>
+                    )}
+                    {address && (
+                        <div className="p-6 border-b border-gray-700 last:border-b-0 gap-4 flex items-center">
+                            <MapPinIcon className="w-4 h-4 text-white" />
+                            <p className="label-large text-white">{address}</p>
+                        </div>
```
oleg-semyonov, line 74: Can we use `<address>` tag here for accessibility and semantics?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

packages/ui/src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <Story />
+            </AnalyticsContextProvider>
+        ),
+    ],
+} as Meta<typeof Footer>;
+
+const FooterTemplate: StoryFn<typeof Footer> = (args) => <Footer {...args} />;
+
+export const Default = FooterTemplate.bind({});
+Default.args = {
```
oleg-semyonov, line 30: As with previous PRs, can we get more story variations for all components? We need to make sure that the component looks good when not every field is populated.

packages/ui/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+    hasStandaloneContactsPage,
+    locale,
+    newsroom,
+    publicGalleriesCount,
+}: Props) {
+    const hasSocialMedia = hasAnySocialMedia(companyInformation);
+
+    return (
+        <footer className={className}>
+            <div className="p-12 flex flex-col md:flex-row bg-gray-900 justify-between">
```
oleg-semyonov, line 38: This is partly a design question, but shouldn't the header and footer contents have a max-width? (i.e. wrapped in a container)

packages/ui/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+                <div className="flex flex-col gap-6">
+                    {hasSocialMedia && (
+                        <SocialMedia
+                            iconClassName="text-white"
```
oleg-semyonov, line 42: Icons are missing hover styling

packages/ui/src/components/Footer/Footer.tsx
fgyimah: @iamhugofelix can you advice here

packages/ui/src/components/ContactCard/ContactCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+                )}
+                <div className="flex flex-col gap-1">
+                    <h4 className="title-x-small">{name}</h4>
+                    <p className="label-large text-gray-600">{subtitle}</p>
+                </div>
+            </div>
+            <div className="flex items-center flex-wrap gap-4">
+                {email && (
+                    <a
+                        className="label-large text-accent flex items-center gap-1"
```
oleg-semyonov, line 43: Sorry, missed during previous review. These links are also missing interaction styling.

packages/ui/src/components/Footer/Footer.stories.tsx
oleg-semyonov: Hey @fgyimah, I asked to add variations on all components 🤔 
Both ContactCard and Boilerplate can look completely different without particular fields

packages/ui/src/components/ContactCard/ContactCard.tsx
fgyimah: Updated

packages/ui/src/components/Boilerplate/Boilerplate.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Link } from '../Link';
+
+export interface Props {
+    className?: string;
+    companyInformation: NewsroomCompanyInformation;
+}
+
+export function Boilerplate({ className, companyInformation }: Props) {
+    const { about, phone, email, address, website } = companyInformation;
+    const hasContactInformation = hasAnyContactInformation(companyInformation);
```
oleg-semyonov, line 21: I think we should not render anything if there is no about information and no contact information.

oleg-semyonov (approved):
Looks good! I've got one minor comment on the Boilerplate

packages/ui/src/components/Boilerplate/Boilerplate.tsx
fgyimah: Updated

PR #440, opened 2023-11-07
[DEV-11851] Add search modal & results components to storybook

fgyimah:


packages/ui/src/components/StoryCard/StoryCard.tsx
```diff
@@ -36,14 +37,16 @@ export function StoryCard({
...
                 className,
             )}
         >
             <Link
                 className={twMerge(
                     'aspect-[10/7]',
-                    size === 'large' ? 'w-full md:aspect-[27/17]' : 'w-1/3 hidden md:block',
+                    size === 'large'
+                        ? 'w-full md:aspect-[27/17]'
+                        : 'w-1/3 hidden md:block md:max-w-[200px]',
```
MohammadxAli, line 49: One approach that I have seen in many libraries/apps using `twMerge` is to introduce a `cn()` utility which leverages `clsx` or `classNames` within `twMerge`:

```tsx
import { clsx, type ClassValue } from "clsx";
import { twMerge } from "tailwind-merge";

export function cn(...inputs: ClassValue[]) {
  return twMerge(clsx(inputs));
}
```

Then you can still use the good old way of doing styles like we used to do before using the objects presentations.

```tsx
cn('aspect-[10/7]', { "w-full md:aspect-[27/17]": size === 'large' })
```

Bounce point, it's _shorter_ :)

(Somehow a big deal for a utility which gets reused a lot)

What do you think Francis?

fgyimah: 👍🏾 I do agree, and it keeps it consistent with how we've been doing it earlier, I would leave it for later though as we'll have to change most parts where we use this syntax

packages/ui/src/components/SearchModal/SearchModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: Props) {
+    const { ALGOLIA_API_KEY, ALGOLIA_APP_ID, ALGOLIA_INDEX } = algoliaConfig;
+
+    const searchClient = useMemo(
+        () => algoliasearch(ALGOLIA_APP_ID, ALGOLIA_API_KEY),
+        [ALGOLIA_API_KEY, ALGOLIA_APP_ID],
+    );
+
+    return (
+        <Dialog.Root open={isOpen} onOpenChange={onOpenChange}>
```
oleg-semyonov, line 48: ❌ 

The modal can't be closed on mobile 😕

oleg-semyonov (changes requested):
Looks good, but I spotted one major issue on mobile

packages/ui/src/components/SearchModal/SearchModal.tsx
fgyimah: https://github.com/prezly/theme-kit-js/pull/440/commits/99f78ceb1773d56de7efe17a3fc63ce2a45a4001

packages/ui/src/components/SearchModal/components/SearchBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+                className="w-full border border-gray-200 rounded px-3 py-2"
+                icon={isMd ? MagnifyingGlassIcon : undefined}
+                inputClassName="p-1 focus:outline-transparent focus-within:outline-transparent"
+                name="query"
+                onChange={(event) => setQuery(event.currentTarget.value)}
+                // TODO: add translations
+                placeholder={`Search ${newsroomName}...`}
+                type="search"
+                value={input}
+            />
```
MohammadxAli, line 51: When search input is focused, there is an ugly blue `outline` around it. Tested in Firefox:

image

Available when element state is `:focus-visible`.

It probably makes sense to apply `focus-visible:outline-0` to `Input` component.

MohammadxAli (approved):
Well done 👍 Just a small remark that would be nice to address 👇

packages/ui/src/components/SearchModal/components/SearchBar.tsx
fgyimah: https://github.com/prezly/theme-kit-js/pull/440/commits/6241cd1541586721a9d8626d1bb92d38dcaa4383

PR #442, opened 2023-11-10
[DEV-11955] & [DEV-11956] Added Featured Media and Category components

fgyimah:


packages/ui/src/components/CategoryBlock/CategoryBlock.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const { name, description } = getLocalizedCategoryData(category, localeObj);
+
+    return (
+        <div
+            className={twMerge(
+                'flex flex-col p-6 w-full md:w-64 rounded border border-gray-200',
+                className,
+            )}
+        >
+            <p className="subtitle-small">{name}</p>
```
oleg-semyonov, line 25: IMO we should use heading tags for card titles to improve semantics. An `h3` or `h4` would work well here.

packages/ui/src/components/GalleryMedia/GalleryMedia.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <Link
+            href={`/media/album/${uuid}`}
+            className={twMerge('relative w-full md:w-max group', className)}
+            contentClassName="p-0"
+            localeCode={locale}
+        >
+            {cover && (
+                <UploadcareImage
+                    className={twMerge(
+                        'aspect-[17/12] md:aspect-[6/7] w-full max-w-full md:max-w-[258px]',
```
oleg-semyonov, line 29: 258px sounds a bit too specific. Wouldn't 256px work here, so that we can use `md:max-w-64` class name and adhere to the grid?

fgyimah: Sure will update, used the width from the design system that's why I ended with this

fgyimah: Had to use `md:max-w-[256px]` here as max-width does not use the usual size variations of tailwind

PR #454, opened 2023-11-13
[DEV-11947] Switch `theme-kit-js` to `pnpm`

fgyimah:


.github/workflows/lint.yml
```diff
@@ -16,20 +16,28 @@ jobs:
...
+            - uses: actions/cache@v3
+              name: Setup pnpm cache
+              with:
+                path: ${{ env.STORE_PATH }}
+                key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
                 restore-keys: |
-                  ${{ runner.os }}-build-${{ env.cache-name }}-
-                  ${{ runner.os }}-build-
-                  ${{ runner.os }}-
+                  ${{ runner.os }}-pnpm-store-
```
e1himself, line 37: We have a reusable Github action to reduce repetitive code in such scenarios:

https://github.com/prezly/setup-github-actions/

pnpm-lock.yaml
```diff
@@ -0,0 +1,16622 @@
+lockfileVersion: '6.0'
```
e1himself, line 1: I suggest to ignore this file and not commit it to the repo. 

The reasoning I'm basing this suggestion on can be found here: 
https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

PR #457, opened 2023-11-14
[DEV-11962] Storybook components QA

fgyimah:
Storybook link: https://6530e13a6911737a3a258bcc-ovdfnvvgnt.chromatic.com/

### DIFF
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -19,12 +19,9 @@ export function Button({
     rounded,
     ...buttonProps
 }: Button.Props) {
-    const iconClassName = twMerge(
-        !children && variation !== 'navigation'
-            ? // Adjusted icon height for square buttons
-              'w-4 h-6'
-            : 'w-6 h-6',
-    );
+    const isIconOnly = Boolean(icon && !children);
+    const hasLeftIcon = Boolean(icon && iconPlacement === 'left');
+    const hasRightIcon = Boolean(icon && iconPlacement === 'right');
 
     return (
         <button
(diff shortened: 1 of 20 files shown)


packages/ui/src/components/Button/Button.tsx
```diff
@@ -19,21 +19,20 @@ export function Button({
...
-            ? // Adjusted icon height for square buttons
-              'w-4 h-6'
-            : 'w-6 h-6',
-    );
+    const isIconOnly = Boolean(icon && !children);
+    const hasLeftIcon = Boolean(icon && iconPlacement === 'left');
+    const hasRightIcon = Boolean(icon && iconPlacement === 'right');
 
     return (
         <button
```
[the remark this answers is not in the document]
alpower: Nice one @MohammadxAli - wondering if just a `w-*` would suffice? Taking a look now at the PR.

packages/ui/src/components/Navigation/Navigation.tsx
```diff
@@ -188,8 +203,8 @@ export function Navigation({
                                         target="_blank"
                                         rel="noopener noreferrer"
                                     >
-                                        {new URL(externalSiteLink).hostname}
-                                        <ArrowUpRightIcon className="ml-1 w-2 h-2" />
+                                        {externalLinkLabel}
+                                        <ArrowUpRightIcon className="ml-1 w-4 h-4" />
```
alpower, line 207: @fgyimah I noticed quite some difference between our nav and yours, both in icon size and spacing between them

CleanShot 2023-11-14 at 17 42 30@2x

Note we've used the heroicons mini for the dropdowns and external links - they are smaller but bolder, and you chouls be able to use them at fullsize (20px) without having to resize them. Other standalone icons use the fullsize 24px versions, but thought I'd let you know in case you missed that difference!

CleanShot 2023-11-14 at 17 44 08@2x

alpower (changes requested):
Great job - there was a decent amount of feedback and you've nailed pretty much all of it. Few comments about icon usage and spacing. I'll let @iamhugofelix cast his eye over the PR tomorrow (there's also a design question about chevron consistency for him in figma we need to get right first before feeding back to you)

packages/ui/src/components/Button/Button.tsx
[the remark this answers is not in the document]
MohammadxAli: They're already using a `w-*` class :)

image

Another option is to utilize [`shrink-0`](https://tailwindcss.com/docs/flex-shrink) (`flex-shrink: 0`) on the icon to prevent it from shrinking. This way, a `min-w-*` is no longer necessary.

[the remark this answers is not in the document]
fgyimah: Updated

packages/ui/src/components/Navigation/Navigation.tsx
fgyimah: Updated

MohammadxAli (approved):
Well done @fgyimah 👍

Looks good to me code-wise!

packages/ui/src/components/StoryCard/StoryCard.tsx
```diff
@@ -84,8 +84,9 @@ export function StoryCard({
                     <Link href={`/${slug}`} locale={false}>
                         <h2
                             className={twMerge(
-                                'cursor-pointer',
+                                'cursor-pointer hove',
```
MohammadxAli, line 87:
```suggestion
                                'cursor-pointer',
```

fgyimah: Good catch 👍🏾

PR #492, opened 2023-12-04
[CARE-3532] Allow bypassing locale code shortening in URLs

oleg-semyonov:
I figured that we shouldn't invest in a more complex setup for a one-off custom theme feature. This logic will also likely heavily change with next major Theme Kit version, so I went with this simple (albeit a bit hacky) approach.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
     "access": "public"
   },
   "useWorkspaces": true,
-  "version": "6.6.1-alpha.0"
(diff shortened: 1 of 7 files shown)


oleg-semyonov:
@e1himself I'm merging this since it's a blocker for a Client TODO, but you might want to keep this PR in mind when working on the `next` branch.

e1himself:
Hm... 

What does this change mean? That the theme app will be generating ULRs with full locale codes when the flag is ON?

PR #573, opened 2024-02-06
[DEV-12376] Fix - Valid ESM build

e1himself:
The previous build setup was not working for us, because:

- `vite` build was bundling all files into aggregate chunks (making the `use client` directive impossible)
- `use client` directives were erased by the build system
- the "ESM" build artifacts were not valid ESM modules too

--- 

What's done:

- Replace `vite` build system with `babel`  (which makes it valid ESM/CommonJS)
- Extract css-in-js stylesheets to a separate PostCSS build


MohammadxAli:
Tests are failing...

e1himself:
> Tests are failing...

It's not new. The Chromatic check has been broken for a while already. :see_no_evil:

MohammadxAli (approved):

fgyimah (approved):
👍🏾

PR #593, opened 2024-02-23
[DEV-12530] Feature - Intl pluralization support

e1himself:
https://github.com/prezly/theme-kit-js/blob/2289f20e0ea2ee570a8c48dc42c4dad7c3b1810c/packages/intl/i18n/en.json#L80-L82

### DIFF
--- a/packages/intl/i18n/en.json
+++ b/packages/intl/i18n/en.json
@@ -78,10 +78,16 @@
     "message": "Something went wrong. Please try submitting the form again"
   },
   "mediaGallery.imagesCount": {
-    "message": "{imagesCount, plural, =0 {No images} one {# image} other {# images} }"
+    "message": "{imagesCount, plural, =0 {No images} one {# image} few {# images} other {# images} }"
   },
   "mediaGallery.title": {
     "message": "Media galleries"
+  },
+  "mediaGallery.titleSingular": {
+    "message": "Media gallery"
+  },
+  "mediaGallery.videosCount": {
+    "message": "{videosCount, plural, =0 {No videos} one {# video} few {# videos} other {# videos} }"
   },
   "misc.embargoMessage": {
     "description": "Embargo means that the story is not allowed to be shared before the specified date",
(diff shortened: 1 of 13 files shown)


packages/intl/i18n/en.json
```diff
@@ -77,9 +77,18 @@
...
+    "message": "{imagesCount, plural, =0 {No images} one {# image} few {# images} other {# images} }"
+  },
   "mediaGallery.title": {
     "message": "Media galleries"
   },
+  "mediaGallery.titleSingular": {
+    "message": "Media gallery"
+  },
+  "mediaGallery.videosCount": {
+    "message": "{videosCount, plural, =0 {No videos} one {# video} few {# videos} other {# videos} }"
```
kudlajz, line 90: I've added the `few` variant in there as well, as I know Czech uses that for example, but now it makes me think if we should add the other ones as well, like `many`? 🤔 @e1himself 

Or if we should omit the `few` as well and just use `0`, `one` and `other`? With just these three, it's gonna be wrong in some languages.

e1himself: Hm... I was thinking that this string is for English language, while other languages would define their own `few` `many` and `other` strings.

e1himself: For English we only need `zero`/`=0`, `one` and `other`.

kudlajz: Hmm, yeah that makes sense, just wondering if the Crowdin translators will know they should do that? I don't have experience with translating plural strings 😄

e1himself: I don't know this either :man_shrugging: 

Just to be safe, we can go all-in and define all possible versions in the source English string to make sure translators know about all available versions for translation.

I.e. let's add `many` to the list?

kudlajz: Yeah, that was my intention 👍 I'm just worried if it's gonna drive the bill up if some languages only use two forms (similar to English), but the word has to be translated like 3-4 times 🤔

e1himself: I wouldn't worry too much about the bill. It's better to safe time and make less mistakes. It's just a few words to translate after all :)

kudlajz (approved):
Great 👍

PR #634, opened 2024-03-26
[CARE-4368] Display UTC offset when formatting time

kudlajz:
`FormattedTime` will now return the full GMT offset and GMT will be replaced by UTC, so the result will be something like this `08:00 AM UTC+01:00`

### DIFF
--- a/packages/react/src/intl/BaseFormattedTime.tsx
+++ b/packages/react/src/intl/BaseFormattedTime.tsx
@@ -14,7 +14,7 @@ export function BaseFormattedTime({
 
     return (
         <time {...attributes} dateTime={dateTime.toISOString()}>
-            {formatTime(dateTime, { locale, timezone })}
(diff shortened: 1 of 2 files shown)


packages/react/src/intl/BaseFormattedTime.tsx
```diff
@@ -14,7 +14,7 @@ export function BaseFormattedTime({
 
     return (
         <time {...attributes} dateTime={dateTime.toISOString()}>
-            {formatTime(dateTime, { locale, timezone })}
+            {formatTime(dateTime, { locale, timezone }).replace('GMT', 'UTC')}
```
kudlajz, line 17: Do you think there's a better way to do this?

e1himself: I am not aware of a better way. This is good enough :)

e1himself (approved):

PR #671, opened 2024-05-09
[CARE-4942] Bump @prezly/sdk

e1himself:
Bump the following dependencies to their latest versions:

- @prezly/sdk
- @prezly/content-renderer-react-js
- @prezly/analytics-nextjs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,13 +7,6 @@ jobs:
     name: Run tests
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        # The first item in the matrix should be the minimal supported SDK version, the last item should be the latest supported version.
-        # You should include any major releases that could potentially break the compatibility in between.
-        # Please keep this in sync with the peer dependency version and with the typecheck.yml workflow.
-        sdk-version: [ 20 ]
-
     steps:
       - name: Check out Git repository
         uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


packages/ui/package.json
```diff
@@ -78,8 +78,8 @@
   },
   "dependencies": {
     "@heroicons/react": "^2.0.18",
-    "@prezly/analytics-nextjs": "^1.13.0",
```
e1himself, line 81: I'm surprized to see how outdated this dependency was. :open_mouth:

.github/workflows/test.yml
```diff
@@ -7,13 +7,6 @@ jobs:
     name: Run tests
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        # The first item in the matrix should be the minimal supported SDK version, the last item should be the latest supported version.
-        # You should include any major releases that could potentially break the compatibility in between.
-        # Please keep this in sync with the peer dependency version and with the typecheck.yml workflow.
-        sdk-version: [ 20 ]
```
e1himself, line 15: I've dropped it as we currently support only one version of the SDK dependency.

packages/ui/src/components/ContentRenderer/components/RichText/Link.tsx
```diff
@@ -8,18 +7,12 @@ interface Props {
 }
 
 export function Link({ node, children }: Props) {
-    const { track } = useAnalytics();
     const { href } = node;
 
-    function handleClick() {
-        track(STORY_LINK.CLICK, { href });
-    }
```
e1himself, line 16: Apparently, we don't track this even on the Link level anymore, as there is no constant for it in the latest `@prezly/analytics` package. 

We only track outbound links (leading to other websites) from [the global listener](https://github.com/prezly/analytics/blob/d7cc7cd20829f4e1da9321b4f3822c712ea8e62a/packages/analytics-nextjs/src/components/Analytics.tsx#L20-L40) defined in the Analytics package.

MohammadxAli (approved):

PR #682, opened 2024-05-14
[CARE-4999] Feature - Allow iframing the theme in Prezly app

kudlajz:


packages/nextjs/src/next-config/index.ts
```diff
@@ -38,7 +38,7 @@ export function createNextConfig({ recommendedHeaders = true }: Partial<Options>
                             },
                             {
                                 key: 'Content-Security-Policy',
-                                value: 'upgrade-insecure-requests; report-uri https://csp.prezly.net/report;',
+                                value: "upgrade-insecure-requests; report-uri https://csp.prezly.net/report; frame-ancestors 'self' https://rock.prezly.com https://rock.*.prezly.net http://rock.prezly.test",
```
kudlajz, line 41: I see the docs mention that the parent must match the protocol, but we don't have `https` locally for Prezly 🤔

camilb: @kudlajz Is this necessary to be set here? I thought was enough for preview URLs

kudlajz: I'm not sure what you mean.. we need to be able to iframe the newsrooms in the production app as well 🤔

camilb: Ok, didn't know that. It is alright then.

camilb (approved):
👍

PR #683, opened 2024-05-14
[CARE-4999] Feature - Update CSP header to allow iframing in the app

kudlajz:
This is the same change as in https://github.com/prezly/theme-kit-js/pull/682 but for the older version 7 (still used in Lena, Greta and Marcel themes).


camilb (approved):
👍

PR #805, opened 2024-10-30
[DEV-13725] Fix - Reduce transactions sample rate to 0.5%

kudlajz:
This is the same fix but for the `7.x` version.


mohammadxali (approved):
👍

fgyimah (approved):

PR #811, opened 2024-11-05
Fix - Filter stories by tag name

fgyimah:


packages/core/src/content-delivery/ContentDelivery.ts
```diff
@@ -264,7 +264,7 @@ export function createClient(
                     [`locale`]: localeCode ? { $in: [localeCode] } : undefined,
                     [`status`]: { $in: [Story.Status.PUBLISHED] },
                     [`visibility`]: { $in: [Story.Visibility.PUBLIC] },
-                    [`tag_names`]: tags?.length ? { $in: [tags] } : undefined,
+                    [`tag.name`]: tags?.length ? { $any: [tags] } : undefined,
```
yuriyyakym, line 267: What's the difference between `any` and `in` in case of search predicate?

yuriyyakym: Would `tags?.length ? { $in: tags } : undefined` also work?

yuriyyakym (approved):

mohammadxali (approved):

packages/core/src/content-delivery/ContentDelivery.ts
fgyimah: This particular filter only supports `$any`, `$some` and `$none` filters, $in is not supported

yuriyyakym: But is there any difference between those predicates? 🤔

fgyimah: I'm not sure there is

PR #864, opened 2025-03-13
[DEV-19820] Fix - Special chars in slug matching

yuriyyakym:


e1himself (approved):
It's good that this is solved on the theme level. 
But I believe we also have to fix the backend to replace `'` and `!` in the slugs.

e1himself:
> It is said in issue description that back-end supports both characters. Do we need to change anything on back-end?

I believe it is actually accidental that the backend allows those characters.

PR #989, opened 2025-09-03
[DEV-21506] Add missing translation strings for the Share block

e1himself:
DEV-21506

### DIFF
--- a/packages/intl/i18n/en.json
+++ b/packages/intl/i18n/en.json
@@ -216,6 +216,18 @@
   "serverError.title": {
     "message": "Something went wrong..."
   },
+  "share.copyLink": {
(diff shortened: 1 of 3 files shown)


packages/intl/i18n/en.json
```diff
@@ -216,6 +216,18 @@
   "serverError.title": {
     "message": "Something went wrong..."
   },
+  "share.copyLink": {
+    "message": "Copy link"
```
kudlajz, line 220: We already have `actions.copyShareUrl` so maybe we could re-use it if it's already translated?

I was also thinking if it would make sense to name these `actions.<name>` as these texts are pretty generic.

PR #1004, opened 2025-09-23
[DEV-21708] Feature - Biome

yuriyyakym:
Screenshot from our CI:
image


digitalbase:
Do you know what the speed was before?

kudlajz:
Looks like it was 24 seconds according to this PR: https://github.com/prezly/theme-kit-js/actions/runs/17882270298/job/50851477191

yuriyyakym:
> @yuriyyakym We should probably rename the step in Github action as well since it's not using ESLint anymore 🙂

Good catch. Done: 60c2a289

